PR #434, opened 2015-01-04
Add control over terminal reaction to bells

spresse1:
This patch adds settings to allow control over how the terminal reacts to bells (\a).  Two checkboxes are added to the General section of the General tab - "Play system alert sound on bell" and "Flash terminal on bell".  Previous versions of guake have ignored all bells.  Many other terminals (gnome-terminal) included allow control over bells and some users (myself included) care about the availability of the bell function.
These settings are default-disabled so that behavior does not change for users.

I'm open to making any changes required, please let me know.

Steve


Stibbons:
Merged. Thanks a lot !

PR #435, opened 2015-01-05
fix non-pep8-compliant spacing.  Followup to #434

spresse1:
My previous pull request (#434) causes a Travis CI build failure due to non-pep8 compliant spacing in src/guake/prefs.py.  This pull request fixes that.  Passes ./valdate.py.

Sorry about that.

Steve


Stibbons:
Great :)

PR #439, opened 2015-01-07
Fix for toolbar visibility in full screen mode.

manoevis:
Basically there is a typo for gconf key reference where '/' is missing at the beginning.


Stibbons:
perfect!

PR #461, opened 2015-02-12
Add keybindings to move current tab

Kaian:
Add two keybindings for moving current tab.


Kaian:
I'm not really sure why travis report failure. None of the pep8 errors belong to the changes in the commit. 

Running validate.sh in my machine with pep8 1.5.7 gives no error.

Stibbons:
travis is broken, I'll update it asap

Stibbons:
thx for the commit :)

pztrick:
After pulling in git `master` I'm receiving this error a bunch because I do not have the hotkey set. It appears as "Disabled". Is this the behavior we want or should 1) the install script set the hotkey default or 2) should Python catch the exception?

```
  File "/usr/local/lib/python2.7/dist-packages/guake/gconfhandler.py", line 345, in reload_accelerators
    self.load_accelerators()
  File "/usr/local/lib/python2.7/dist-packages/guake/gconfhandler.py", line 378, in load_accelerators
    key, mask = gtk.accelerator_parse(gets('move_tab_left'))
TypeError: Gtk.accelerator_parse() argument 1 must be string, not None
```

These are the steps I used to re-install:

```
sudo make uninstall
./configure && make && sudo make install
```

cc @Kaian

Kaian:
Not sure, but maybe this could be fixed by adding a default value for those two bindings in the schema? I can create another PR with this change if you want.

kaian/guake@a9f309676cdbcb150bdbeb3c430829254ecc4c8d

Kaian:
This seems to be fixed in `master` 0ea78739a9a95c150c8afac6626371bc24d4ddc1. I remove the previous mentioned commit. cc @pztrick

PR #509, opened 2015-04-27
use dconf for ubuntu launcher hide/width detection

scinart:
The affected code is tested on my Ubunut 15.04 and documented.

current guake's detection about ubuntu launcher's hide state and icon size is not working.
Now dconf is used to solve this problem. (for ubuntu >=12.10)

I believe the change is quite straightforward.

### DIFF
--- a/src/guake/guake_app.py
+++ b/src/guake/guake_app.py
@@ -26,9 +26,12 @@ import gconf
 import gobject
 import gtk
 import os
+import platform
+import subprocess
 import pygtk
 import sys
 import xdg.Exceptions
+
 
 from urllib import quote_plus
 from urllib import url2pathname
@@ -693,21 +696,27 @@ class Guake(SimpleGladeApp):
         window_rect = screen.get_monitor_geometry(monitor)
 
         if os.environ.get('DESKTOP_SESSION') == "ubuntu":
-            unity_hide = self.client.get_int(KEY('/apps/compiz-1/plugins/'
-                                                 'unityshell/screen0/options/launcher_hide_mode'))
-            # launcher_hide_mode = 1 => autohide
-            if unity_hide != 1:
-                # Size of the icons for Unity in Ubuntu <= 12.04
-                # TODO Ubuntu 12.10 use dconf :
-                # /org/compiz/profiles/unity/plugins/unityshell/icon-size
+
+            # For Ubuntu 12.10 and above, use dconf :
+            # For Ubuntu 12.04 and below, use
+            # see if unity dock is hide => unity_hide
+            # and the width of unity dock. => unity_dock
+            if float(platform.dist()[1])>=12.10:
+                unity_hide = int(subprocess.check_output(
+                    ['/usr/bin/dconf', 'read',
+                     '/org/compiz/profiles/unity/plugins/unityshell/launcher-hide-mode']))
+                unity_dock = int(subprocess.check_output(
+                    ['/usr/bin/dconf', 'read',
+                     '/org/compiz/profiles/unity/plugins/unityshell/icon-size']))
+            else:
+                unity_hide = self.client.get_int(KEY('/apps/compiz-1/plugins/'
+                                                     'unityshell/screen0/options/launcher_hide_mode'))
                 unity_icon_size = self.client.get_int(KEY(
                     '/apps/compiz-1/plugins/unityshell/screen0/options/icon_size'))
-                if not unity_icon_size:
-                    # If not found, it should be because of newer implementation of unity.
-                    # Dock is 64 pixel of width on my system, hope this is so on others...
-                    unity_dock = 64
-                else:
-                    unity_dock = unity_icon_size + 17
+                unity_dock = unity_icon_size + 17
+
+            # launcher_hide_mode = 1 => autohide
+            if unity_hide != 1:
                 print("correcting window width because of launcher width {} "
                       "(from {} to {})".format(
                           unity_dock, window_rect.width, window_rect.width - unity_dock))


src/guake/guake_app.py
```diff
@@ -693,21 +696,27 @@ def set_final_window_rect(self):
...
-                # /org/compiz/profiles/unity/plugins/unityshell/icon-size
+
+            # For Ubuntu 12.10 and above, use dconf :
+            # For Ubuntu 12.04 and below, use
+            # see if unity dock is hide => unity_hide
+            # and the width of unity dock. => unity_dock
+            if float(platform.dist()[1])>=12.10:
+                unity_hide = int(subprocess.check_output(
+                    ['/usr/bin/dconf', 'read',
+                     '/org/compiz/profiles/unity/plugins/unityshell/launcher-hide-mode']))
```
pypingou: Oh please no, we do not want to shell out command.

src/guake/guake_app.py
```diff
@@ -693,21 +696,27 @@ def set_final_window_rect(self):
...
-            if unity_hide != 1:
-                # Size of the icons for Unity in Ubuntu <= 12.04
-                # TODO Ubuntu 12.10 use dconf :
-                # /org/compiz/profiles/unity/plugins/unityshell/icon-size
+
+            # For Ubuntu 12.10 and above, use dconf :
+            # For Ubuntu 12.04 and below, use
+            # see if unity dock is hide => unity_hide
+            # and the width of unity dock. => unity_dock
+            if float(platform.dist()[1])>=12.10:
```
pypingou:
```
>>> platform.dist()
('fedora', '21', 'Twenty One')
````
This will go wrong in so many ways :)

File: src/guake/guake_app.py (hunk 1 above)
Comment (scinart) [inline src/guake/guake_app.py]: OK, what is the python way to do it then. I'm a python amateur.

File: src/guake/guake_app.py (hunk 2 above)
Comment (scinart) [inline src/guake/guake_app.py]: Why? Ubuntu version always seem to be a float number.
Please give a more robust fragment of code then. Because it is not working currectly now. 
both `unity_hide` and `unity_dock` get wrong value.

Comment (pypingou) [inline src/guake/guake_app.py]: Well at first you would want to check if you are on ubuntu or not, because ``float(21) > 12.10`` and that can be the case for many other OS.

But in general, I am not in favor of adding distro-specific pieces of code if we can avoid it.

File: src/guake/guake_app.py (hunk 1 above)
Comment (pypingou) [inline src/guake/guake_app.py]: I'm fairly sure dconf has a python API.

File: src/guake/guake_app.py (hunk 2 above)
Comment (pypingou) [inline src/guake/guake_app.py]: Ah, missed the check before for the ``DESKTOP_SESSION``, I actually wonder if we would not want to replace it with a check using ``platform``.

Comment (scinart) [inline src/guake/guake_app.py]: The original code has `if os.environ.get('DESKTOP_SESSION') == "ubuntu":`, which didn't show up in diff view. these changes is under the condition that it is a ubuntu now.

The original code already a platform-specific solution. I'm just making it work. I'd like that you give a better way.

```
         if os.environ.get('DESKTOP_SESSION') == "ubuntu":
-            unity_hide = self.client.get_int(KEY('/apps/compiz-1/plugins/'
-                                                 'unityshell/screen0/options/launcher_hide_mode'))
-            # launcher_hide_mode = 1 => autohide
-            if unity_hide != 1:
-                # Size of the icons for Unity in Ubuntu <= 12.04
-                # TODO Ubuntu 12.10 use dconf :
-                # /org/compiz/profiles/unity/plugins/unityshell/icon-size
+
+            # For Ubuntu 12.10 and above, use dconf :
+            # For Ubuntu 12.04 and below, use
+            # see if unity dock is hide => unity_hide
+            # and the width of unity dock. => unity_dock
+            if float(platform.dist()[1])>=12.10:
+                unity_hide = int(subprocess.check_output(
+                    ['/usr/bin/dconf', 'read',
+                     '/org/compiz/profiles/unity/plugins/unityshell/launcher-hide-mode']))
+                unity_dock = int(subprocess.check_output(
+                    ['/usr/bin/dconf', 'read',
+                     '/org/compiz/profiles/unity/plugins/unityshell/icon-size']))
+            else:
+                unity_hide = self.client.get_int(KEY('/apps/compiz-1/plugins/'
+                                                     'unityshell/screen0/options/launcher_hide_mode'))
                 unity_icon_size = self.client.get_int(KEY(
                     '/apps/compiz-1/plugins/unityshell/screen0/options/icon_size'))
-                if not unity_icon_size:
-                    # If not found, it should be because of newer implementation of unity.
-                    # Dock is 64 pixel of width on my system, hope this is so on others...
-                    unity_dock = 64
-                else:
-                    unity_dock = unity_icon_size + 17
+                unity_dock = unity_icon_size + 17
```

src/guake/guake_app.py
scinart: OK. Then I hope that this issue is fixed the correct way soon.

src/guake/guake_app.py
Stibbons: indeed this block is only for ubuntu

src/guake/guake_app.py
Stibbons: i'll check that but I remember that there was some dependencies problem... so I ended up executing this command for other use case :(

pypingou: arf :( we would really prefer to avoid this.

Stibbons: that's a chicken and egg problem. This should be handled by the packager (for debian, for rehat, ...). But, as a User, I want to be able to use Guake directly from source. And as a Maintainer, I don't want packager to introduce regression in my software.... So I don't see how to handle distribution specific stuff differently.

Maybe a good thing to do is to move this trick in a spefic module so distribution hack are better identified

Stibbons:
I merge and give a try on my system. I'll switch to ubuntu 15.04 today

Issue #473, opened 2015-03-04
Guake Turns Gray sometimes


bchoi78:
I don't know quite how to replicate the bug but sometimes my guake terminal turns solid gray and I can't access it. I know at least CTRL+C works (can close an open python GUI doing so) but I can't see anything.

Seems to happen when guake opens/closes the terminal too quickly (like when the GUI runs)

Stibbons:
indeed, i ve seen some freeze from time to time :(

Stibbons:
which window environment are you using?

Meistache:
I've noted this problem when I move a tab on a position, then I close it. Whatever the tab which gets that position blanks, even if I close another one or move another tab into that position, the next in that position will suffer the same way.

SparkyBluefang:
I can confirm what Meistache describes. I definitely see this happening after rearranging tab and then closing it. The scroll bar for the affected tab remains, but it looks like the terminal area is collapsed or removed.

![guake_tab](https://cloud.githubusercontent.com/assets/3150882/7221430/d565d0d2-e6b8-11e4-9a54-d5fe1927c6ef.png)

I'm using xfwm4 4.12.2

Seraff:
This is becase Guake kills a wrong terminal when rearranged tab is closed. PR #512 fixes this issue.

mart-r:
Hey!
I've got a similar - yet not the same - problem.
For some reason my guake turned completely gray. Unlike the picture previously posted, my entire thing went gray (the tabs as well).
![guake_bugged](https://cloud.githubusercontent.com/assets/15647757/10933497/85e7a18c-82e1-11e5-8eb9-f4283f64ce7f.png)

I'm running 2 python scripts on there for close to 11 hours now, both localized to one of my 4 cores, they seem to work fine (won't know for sure until they're done, maybe another 10 hours), they're solving PDEs so it's calculation heavy. 
I'm also running the system monitor (seen in the background), which kept working just fine.

The only problem is that the whole box is grayed out. It seems I might be able to select stuff under all the gray (the select cursor appears), yet if I don't see what it is it's useless.

I'm fairly certain I didn't do any tab rearranging.

I have Guake v 0.4.4

Thanks for the help.

webhamster:
Same here with Guake 0.8.8

helix84:
[...]

PR #512, opened 2015-05-01
Fixing problems with pasting from clipboard after tabs reorder

Seraff:
Open Guake with 2 tabs. Paste some text in theese terminals, everything is ok.
Swap them. Now when you try to paste text in the first tab, it will be pasted in the second tab.

This is because we don't change order of GuakeNotebook's term_list when the order of GuakeNotebook's pages changes. When we paste something, accel_paste_clipboard() method calls (guake_app.py), it detects terminal in GuakeNotebook by the **index of the current page**.

I redefined reorder_child() method of GuakeNotebook. So when you reorder pages in GuakeNotebook, terminals will reorder automatically.

### DIFF
--- a/src/guake/guake_notebook.py
+++ b/src/guake/guake_notebook.py
@@ -24,6 +24,14 @@ class GuakeNotebook(Notebook):
         # used to kill the process when closing a tab
         self.pid_list = []
 
+    def reorder_child(self, child, position):
+        """ We should also reorder elements in term_list
+        """
+        old_pos = self.get_children().index(child)
+        terms = self.term_list
+        terms[old_pos], terms[position] = terms[position], terms[old_pos]
+        super(GuakeNotebook, self).reorder_child(child, position)
+
     def has_term(self):
         return self.term_list
 


pypingou:
Cool, :+1: for me

Seraff:
Also fixes #473.

Stibbons:
goog catch. All this code is buggy, tab management is buggy, this prevent to easily do something like split terminal or save session... If I had time I would rewrite all of this.

Thanks for the PL and time you spend on debuging this issue!

jellysheep:
This bug as well as #473 are still present when not swapping tabs next to each other.
Steps to reproduce (same on 0.7.2 and current master):
* create 3 tabs (a, b, c)
* move tab c to the left (c, a, b)
* copy/paste in tabs a and b is interchanged
* close tab b
* now tab a is grey as described in #473

This is because you do not only have to swap the terms by `terms[old_pos], terms[position] = terms[position], terms[old_pos]` but also move the terms in between.
I am not familar with the Guake code, but I can try to find a fix.

PR #514, opened 2015-05-04
Be careful with the selection copied

pypingou:
Make sure there was something selected before stripping it

Fixes downstream bug: https://bugzilla.redhat.com/1209759


pypingou:
Note that the problem seems to be occurring quite often: https://retrace.fedoraproject.org/faf/problems/644791/

Stibbons:
good catch

PR #528, opened 2015-05-08
fixed get_tab_count and get_tab_name dbus interfaces (wrong term_list…

Ozzyboshi:
fixed get_tab_count and get_tab_name dbus interfaces (wrong term_list variable)


Stibbons:
merged

PR #530, opened 2015-05-09
added get_gtktab_name dbus interface

Ozzyboshi:
As discussed before making a pull request for the dbus interface i need to develop guake-indicator
Thanks for your help


Stibbons:
merged. thanks for the pl !

PR #546, opened 2015-05-16
prompt_on_close_tab option

tobz1000:
A feature I was missing from gnome-terminal. Like prompt_on_quit, but for individual terminals. Values are "never" (default), "with process running", and "always".

Additionally, if set to "with process running", the quit-prompt will be shown
when quitting Guake if there are any subprocesses running. If set to "always",
it essentially overrides the prompt_on_quit option.

The option is in the GUI as a dropdown, just above prompt_on_quit. prompt_on_quit now behaves more a sub-option of prompt_on_close_tab.

If one of the two prompts is open (close tab or quit Guake) and the other opens, the first will be destroyed. Otherwise I found Guake wouldn't quit until the close-tab prompt was dismissed.

I changed the prompt text slightly; it includes the number of tabs open when quitting.

Please let me know what you think if you get a chance :)

### DIFF
--- a/data/guake.schemas
+++ b/data/guake.schemas
@@ -88,9 +88,21 @@
             <type>bool</type>
             <default>true</default>
             <locale name="C">
-                <short>Prompt when quiting.</short>
-                <long>Ask for confirmation when guiting Guake and there
-                processes running.</long>
+                <short>Prompt when quitting.</short>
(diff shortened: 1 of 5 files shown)


Stibbons:
Look great ! I test it and merge it asap

Stibbons:
Tested and approuved !

PR #547, opened 2015-05-16
Some minor fixes and updated Polish translation

piotrdrag:


src/guake/about.py
```diff
@@ -43,5 +43,5 @@ def __init__(self):
         img = gtk.gdk.pixbuf_new_from_file(ipath)
         dialog.set_property('logo', img)
 
-        dialog.set_name('Guake Terminal')
+        dialog.set_name(_('Guake Terminal'))
```
Stibbons: beware this won't work, you need to include '_' (https://travis-ci.org/Guake/guake/builds/62825994)

Stibbons:
Thanks a lot for all your contributions :)

Stibbons:
I have a small comment on one patch, overall it's good

src/guake/about.py
piotrdrag: Do you mean adding "from guake.common import _"?

Stibbons: I guess so, this is an import that needs to be added in each file that use '_'

piotrdrag: I'm afraid I don't know anything about programming. Let's see if this works.

Stibbons: let's merge!

piotrdrag:
Is there a reason why this wasn't merged?

Stibbons:
Sorry, I closed without clicking on merge. Your change should have been integrated into c6e800a88a9086a671684f3ecd86873598797269. Can you check if all your translations are sync in pl.po?

PR #559, opened 2015-06-02
set default value for /apps/guake/general/debug_mode

ohaessler:
The default value for /apps/guake/general/debug_mode is not set, which gives a warning while installing it from an RPM on Fedora and RHEL.


Stibbons:
thanks!

PR #560, opened 2015-06-03
Replaced GlobalHotkeys with the keybinder module

### LINKED ISSUE #510: replace globalhotkey by keybinder
Url: https://github.com/engla/keybinder
Used by Final Terminal.

Seem more futur proof than our globalhotkey package. Give a try to see if it fixes issue with key binding

### LINKED ISSUE #558: Guake fails to start due to a GlobalHotkey related C call
This is the traceback:

      File "/usr/lib/python2.7/runpy.py", line 162, in _run_module_as_main
        "__main__", fname, loader, pkg_name)
      File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
        exec code in run_globals
      File "/usr/lib/python2.7/site-packages/guake/main.py", line 235, in <module>
        exec_main()
      File "/usr/lib/python2.7/site-packages/guake/main.py", line 231, in exec_main
        if not main():
      File "/usr/lib/python2.7/site-packages/guake/main.py", line 147, in main
        instance = Guake()
      File "/usr/lib/python2.7/site-packages/guake/guake_app.py", line 287, in __init__
        self.hotkeys = guake.globalhotkeys.GlobalHotkey()
    SystemError: NULL result without error in PyObject_Call

This started happening since today. I tried the usual reboot-and-hope-things-work scenario, but didn't work.
For the record, I am using Arch Linux.

jenrik:
This PR fixes #558 and closes #510
Note: when setting a global hotkey the setter is responsible for calling unbind on the previously bound key. Keybinder doesn't have a unbind_all function.


Stibbons:
Look great!
I tried to reproduce your bug today and completely destroyed my Python install...
So thanks a lot for this fix ! I ll try it tomorrow and merge it asap for next release. This worth jumping to v0.8.0 :)

pypingou:
Just a note to say that the last commit in this project is from June 20*12*, it does not look so maintained anymore :(

Stibbons:
I don't think the last commit on globalhotkey is more recent. At least this is an official project already used by other software (such as finalterm). But this need to have an extensive testing of course. I ll simply disable globalhotkey for the moment, to switch back easily if needed.

pypingou:
> At least this is an official project already used by other software (such as finalterm)

Official project not touched for 3 years?

Is finalterm using it or is it bundling it and has patched it all over the places? If it is the later, this might end up being more work for us at the end (plus all the issues that bundling something bring)

src/guake/gconfhandler.py
```diff
@@ -364,7 +364,6 @@ def reload_globals(self, client, connection_id, entry, data):
         method. If more global hotkeys should be added, just connect
         the gconf key to the watch system and add.
         """
-        self.guake.hotkeys.unbind_all()
```
Stibbons: Note: there is a method keybinder_unbind_all (https://github.com/engla/keybinder/blob/master/libkeybinder/bind.c)

jenrik: Yes, but it isn't exposed in the python interface

Stibbons:
Sad to see FinalTerm is discontinued... Look like guake is one of the few hotkey terminal for linux left on the field...
Anyway, I'll take care of your PL from now, intensively test it before merging it.

I hope "libkeybinder0" is indeed packaged with all distribution...

ivoshm:
With Manjaro/Arch is bundled libkeybinder2 (https://github.com/engla/keybinder) and libkeybinder3 (https://github.com/engla/keybinder/tree/keybinder-3.0) too. 

R.I.P. FinalTerm http://worldwidemann.com/finally-terminated/

src/guake/gconfhandler.py
Stibbons: ok. Does it has any consequence?

jenrik: It will be a problem if their is gonna be more places in Guake where you can change your keybindings, because they will all have too have code that unbind the previous key. But we could bundle a modifies python binding that has the unbind_all function exposed in python. If you interested I would gladly try to implement.

Stibbons: it is just a matter of manually call unbind(key) for each registered key. We can make a wrapper in guake that handles this. If you want to do, feel free!

NEWS
```diff
@@ -18,6 +18,7 @@ Version 0.7.3 (dev)
    - grunt (gulp is way better, and faster)
 
   Bug fixes:
+   - Fixes #558 by replacing GlobalHotkeys with keybinder
```
Stibbons: feel free to add your credit ("thanks @jenrik!")

Stibbons:
Merged. I'll try it from now, so far it works like a charm :)

ElijahLynn:
There is a bug with keybinder and the shift key, see issue here for resolution. I am still figuring out how to test it so just crosslinking for now. 

https://github.com/engla/keybinder/issues/7

PR #562, opened 2015-06-06
Global hotkey setters is no longer responsible for calling unbind on previously bound key

jenrik:
When changing a global hotkey the previously bound key is now automatically unbound.

### DIFF
--- a/src/guake/gconfhandler.py
+++ b/src/guake/gconfhandler.py
@@ -11,12 +11,13 @@ import subprocess
 from pango import FontDescription
 from xml.sax.saxutils import escape as xml_escape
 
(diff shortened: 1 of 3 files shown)


src/guake/gconfhandler.py
```diff
@@ -345,8 +346,15 @@ def __init__(self, guake):
         self.client = gconf.client_get_default()
 
         notify_add = self.client.notify_add
-        notify_add(GKEY('show_hide'), self.reload_globals)
```
Stibbons: why do you change it?

Stibbons:
tested. Seems to work perfectly

Issue #564, opened 2015-06-07
a-la guake-indicator custom commands


Stibbons:
Feature: allow users to define there own commands they often use
This can work like guake-indicator, but not with the same level of customization. The source code is heavily using C code, and reimplement lot of trivial stuff in python world (json parsing, GTK event handling,...)

Inserting guake-indicator custom feature in guake has the following advantage:
- compatible with the "hide on focus loss" mode
- automatically available to the user

I also tend to prefere having a custom xml/json file, so the user can back it up easily, vs an import/export feature that is too heavy to use

Screenshot: http://guake-indicator.ozzyboshi.com/screenshots.html
Related issue: #527

@Ozzyboshi what do you think about this?

Ozzyboshi:
I am very glad that Stibbons deems Guake Indicator a useful and valuable project.
I can rewrite or help rewriting the core of guake-indicator in python, we can take advantage of the current Guake "app indicator" to show some custom commands read from an xml or json configuration file.
Are you planning to let the users write their own xml/json file with a text editor?
We could add a tab on the Guake preferences window to provide a GUI interface but this would be more complex.

Another faster and trouble-free way to achieve this could be to merge a light version of guake indicator into the guake package or just advertise guake-indicator within guake (a link/button in the preferences window?)

Unfortunately when I wrote Guake Indicator I didn't think about the possibility to merge my code inside Guake so I used the C language (my favourite). Furthermode Guake indicator is full of extra stuff customized for my own needs and it's very unlikely that Guake users need all the features.

Right now I am improving the scrollbar management, in my opinion it's a more useful feature, after that we could start developing the custom command features if you want.

Stibbons:
I wanted the custom commands to be directly accessible in the contextual menu inside the terminal. For example, I use "hide on focus loss", it is very useful to hide guake when a command is launched, but you cannot click on something in the indicator area and keep guake open.

Also, people that doesn't use ubuntu might also need such commands, even when they don't have an indicator aera (indicator = ubuntu). Will it work on other platform straight forward (fedora, xfce,...) ?

If you manage to integrate your indicator inside the guake icon (to have only 1 icon), it could be a good start.

The things is users (and I) have some "common" commands to execute quite often. I heavily use "custom start script", but having a menu with the custom command display on an hotkey (Ctrl+Alt+b for instance) right inside the terminal would be a killer feature.

[...]

PR #572, opened 2015-06-18
Merge for a-la guake-indicator custom commands #564

Ozzyboshi:
Hello please consider my pull request,this should solve Guake issue #564 
Here what i've done:

1. added new gconf setting for storing config file URI (and not path as previously discussed)
2. added new button on the prefs window that allow the user to select the config json file
3.every time the user changes the file guake will reload the context menu with new commands

for now the user has to edit manually the json file, you should publish the instruction to build it somewhere.
It is very easy to build one: just create a json file witjh  an array named cmds, example

{
 "cmds":[ "ls","df","ping google.it","lpass login","lpass_logout","df -h" ]
}

I am considering to change it and put an array of objects, one field for the context menu description and another with the real command to send to guake, let me know what you think.

I'd like to put a new button or label on the prefs window next to the  select file command to advertise guake-indicator, do you think it is possible?

### DIFF
--- a/data/guake.schemas
+++ b/data/guake.schemas
@@ -987,5 +987,20 @@
             </locale>
         </schema>
 
+        <schema>
+            <key>/schemas/apps/guake/general/custom_command_file</key>
+            <applyto>/apps/guake/general/custom_command_file</applyto>
+            <owner>guake</owner>
+            <type>string</type>
+            <default>~/.config/guake/custom_command.json</default>
+            <locale name="C">
+                <short>Path to the default custom command json file.</short>
+                <long>Path to the default custom command json file.
+                If is blank or the json file is not in a proper format
+                the terminal context menu will be built without custom commands.
+                If the file is a valid json with recognized structure theterminak
+                context menu will be buit with the commands read inside.</long>
+            </locale>
+        </schema>
     </schemalist>
 </gconfschemafile>
--- a/data/prefs.glade
+++ b/data/prefs.glade
@@ -524,24 +524,6 @@ Always</property>
                                       </packing>
                                     </child>
                                     <child>
-                                      <widget class="GtkCheckButton" id="use_vte_titles">
-                                        <property name="label" translatable="yes">Use VTE titles for tab names</property>
-                                        <property name="visible">True</property>
-                                        <property name="can_focus">True</property>
-                                        <property name="receives_default">False</property>
-                                        <property name="use_action_appearance">False</property>
-                                        <property name="use_underline">True</property>
-                                        <property name="active">True</property>
-                                        <property name="draw_indicator">True</property>
-                                        <signal name="toggled" handler="on_use_vte_titles_toggled" swapped="no"/>
-                                      </widget>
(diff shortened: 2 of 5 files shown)


Ozzyboshi:
news for this pull request? Should we close it?

Stibbons:
I forgot it. Indeed, I need to have a look at it. Need a few days for review...

Stibbons:
Merged. I would have preferred to be in a submenu. Some bugs left to fix, I'll try to find time to check it next week. Thanks!

Ozzyboshi:
The submenu implementation should be easy.
If you report me some bugs I try to fix it.
Actually I am thinking about making a new field in the json file (let's say a description field) that will contain the description of the command and the menuitem takes his content from this new field rather than the command, that should prevent very long menuitems if the command is long.

Stibbons:
Yes. Could even be hierarchical:


{
"cmds":[ {description: "ls", cmd: ["ls"]}", ...
}

If the cmd themselves are list it is better (actually safer). If we can get rid of the first cmds level are set the list as root of the Json it will be even easier

Stibbons:
{
"Cmds":[ {"type": "menu", "description": "common commands", "items": [ {"description": "current dir", cmd:["pwd"]]}]
}

Ozzyboshi:
Ok I'll try to write something next week if I have a chance

Stibbons:
I have update my previous command, json is valid so we can do directly the list of items

Ozzyboshi:
Hello Stibbons, I need some hints about your fix after merge (commit 4e85d5368d79f49d3cdce73cc84eb3be88dd5ac1 authored 6 days ago).

The question is about the method def read_custom_commands_file(self) on src/guake/guake_app.py

you turned (my old row 377)
data_file = urllib2.urlopen(self.client.get_string(KEY('/general/custom_command_file')))
into
 with open(file_name) as f:

The problem here is what file_name is, in my ubuntu 14.04 it contains a file URI that is produced when the user selects ths json file in the prefs windows, so it should contain something like : file:///home/user/config.json

file method requires a file name and not a file URI, that's why on my machine the open fails and the custom commands are not shown within my context menu.

So I'd like to ask you if the open(file_name) works in your machine and if yes what OS are you using.
I used a URI rather than a path because I wanted to make this feature as portable as possible across multiple OS and a URI seemed a good choice to me.

Please let me know and sorry if the question is so silly but maybe I am missing something...

Stibbons:
Indeed, I changed it because it doesn't make sense to force usage of a url.
I don't think you want to place this file on a remote server, while it
could be done.
It is also much easier to read a file with python ("with open() as f"), and
this is the portable way (work everywhere). I know that the urllib has
changed with python 3. And using request is better for network connection
that urllib.

I don't have aligned it with the behavior of the file selector, I forgot it.
I think there is a way to get the full path from the file selector and not
the file:// uri.

-----
Gaetan


2015-08-27 10:01 GMT+02:00 Alessio Garzi <notifications@github.com>:

PR #579, opened 2015-07-04
Spixel color palette

dorian-marchal:
I created a color palette, if you're interested, you can add it to Guake by merging this pull request :+1:

Spixel :
spixel

Compared to Tango :
tango


Stibbons:
great!

PR #583, opened 2015-07-09
Fixed typo on Man page

jschwanebeck:
guake.1

Line 20: Dioplay -> Display

### DIFF
--- a/guake.1
+++ b/guake.1
@@ -1,0 +1,64 @@
+.TH "guake" "1"
+.SH "NAME"
+guake \(em A drop-down terminal for GNOME environment.
+.SH "SYNOPSIS"
+.PP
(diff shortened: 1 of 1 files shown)


pypingou:
Did you fix the right file?
I'm seeing the change as if you are adding a whole new file, instead of just patching this one: https://github.com/Guake/guake/blob/master/data/guake.1

Stibbons:
ok look good

pypingou:
So we now have 2 man pages?
https://github.com/Guake/guake/blob/master/data/guake.1
https://github.com/Guake/guake/blob/master/guake.1
...

Stibbons:
weird indeed

pypingou:
There is nothing weird, see my first comment, @jschwanebeck added a new file instead of correcting the existing one...

Issue #288, opened 2014-03-16
`guake --help` and `man guake` are incoherent


Nicolas01:
`man guake` seems to be outdated.  
Please update the manual and the help to display the same information. It's very confusing.

PR #587, opened 2015-07-17
Remove last reference to --show-hide

hansjoachim:
Hi, I was following #288. Thanks for fixing that issue. I discovered that --show-hide was still mentioned in the summary of available options.


pypingou:
:+1: for me

Issue #433, opened 2015-01-01
Broken tray icon with Appindicators


cl0ne:
Cinnamon 2.4.5 (System Tray applet), Arch Linux, libappindicator 12.10, leftmost icon is Guake:
![2015-01-01 03 07 02](https://cloud.githubusercontent.com/assets/271097/5591425/6a611cc4-9163-11e4-9fd1-c67f97abd4c1.png)

As a quick-and-dirty solution I'm using

``` patch
--- guake_app.py    2015-01-01 03:23:58.251836090 +0200
+++ guake_app.py.orig   2014-11-23 19:36:50.000000000 +0200
@@ -146,7 +146,7 @@
             self.tray_icon.connect('activate', self.show_hide)
         else:
             self.tray_icon = appindicator.Indicator(_("guake-indicator"), _("guake-tray"), appindicator.CATEGORY_OTHER)
-            self.tray_icon.set_icon(pixmapfile("guake-tray.png"))
+            self.tray_icon.set_icon("guake-tray")
             self.tray_icon.set_status(appindicator.STATUS_ACTIVE)
             menu = self.get_widget('tray-menu')
             show = gtk.MenuItem(_('Show'))
```

RedGearPro:
Bump! Manjaro with xfce:
![image](https://cloud.githubusercontent.com/assets/8226605/6351210/91851b42-bc07-11e4-9de7-00b9c6ef5b8a.png)
Left one is guake.

sanderboom:
MATE (GTK2):
![image](https://cloud.githubusercontent.com/assets/5956570/8008997/f37b2c04-0ba5-11e5-81b6-989d3151f446.png)
Most left is ... you guessed it

Jesse-Millwood:
I don't know if this applies here or not but with Xubuntu 14.04 the original guake icon shows up in my Notification area but it does not use the icon for my theme. 
![selection_058](https://cloud.githubusercontent.com/assets/5342237/8678659/b4eca44e-2a23-11e5-87ad-9566f1c41966.png)

I have guake 0.7.2 on Xubuntu 14.04. For me this is just a minor annoyance. Otherwise I could not be happier with guake!

thwarted:
Just throwing some more information out there since I'm investigating this issue right now...

[...]

PR #591, opened 2015-08-01
use full path to tray icon with libappindicator

thwarted:
Resolve issue #433, broken tray icon when using python libappindicator bindings.

According to http://people.canonical.com/~ted/libappindicator/current/AppIndicator.html `set_icon` takes an "icon-name", which presumably is not actually intended to be a filename (?) Hard to tell.  Giving it the full path works.

### DIFF
--- a/src/guake/guake_app.py
+++ b/src/guake/guake_app.py
@@ -167,10 +167,10 @@ class Guake(SimpleGladeApp):
         self.preventHide = False
 
         # trayicon!
+        img = pixmapfile('guake-tray.png')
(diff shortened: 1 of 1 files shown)


thwarted:
I found this example code that doesn't use a full path/filename to the icon file, but just a bare "name", so who knows what the intent is.

http://www.eurion.net/python-snippets/snippet/Create%20an%20Application%20Indicator.html

Stibbons:
ok thanks for this pl

Issue #329, opened 2014-06-06
Bold fonts


jperaltac:
Is it possible to add no-bold font option in preferences?

PR #603, opened 2015-08-24
Add option to allow/disallow bold font

helix84:
closes Guake/guake#329


helix84:
Can be tested using:
echo -e "Normal \e[1mBold \e[21mNormal"

Stibbons:
thnk for commit

PR #605, opened 2015-08-28
changed method get_uri in filechooser to get_filename

Ozzyboshi:
gtkfilechooser.get_filename() should return the full path of the file selected (I hope it works coherently across multiple OS like Windows), in this way the  with open(file_name) as f: method you wrote should work.
Please let me know if this is a valid fix for you. Then I'll try to write some code to group the commands within a dedicated menuitem and other stuff we discussed before.

### DIFF
--- a/src/guake/prefs.py
+++ b/src/guake/prefs.py
@@ -408,7 +408,7 @@ class PrefsCallbacks(object):
                                ERASE_BINDINGS[val])
 
     def on_custom_command_file_chooser_file_changed(self, filechooser):
-        self.client.set_string(KEY('/general/custom_command_file'), filechooser.get_uri())
+        self.client.set_string(KEY('/general/custom_command_file'), filechooser.get_filename())
 
 
 class PrefsDialog(SimpleGladeApp):


Stibbons:
This is already done, I don't know why github accept the merge:
https://github.com/Guake/guake/blob/master/src/guake/prefs.py:411

Ozzyboshi:
Did githib automatically merged this pull request without your approval?
How that is possible?Maybe it's because I changed code related to a previously approved PL?
However I just need to know if you agree with this change, now my guake indicator shows the custom commands again.

Stibbons:
Of course I agree, but I already did this very change yesterday!

Ozzyboshi:
In this case it's by bad, this morning I forgot to perform a pull request before working on guake and that's why my pull request has been canceled by github.
Sorry again

Stibbons:
no pbl, but github should have seen it saying it cannot be merged...

The only thing that miss i when you load the preference window, the name of the file is not set. If you can have a look at it, I would help me a lot. I don't think I'll be able to work on guake for some days :)

PR #606, opened 2015-08-31
Improvements for custom commands

Ozzyboshi:
This is a pull request for what we discussed in this thread : https://github.com/Guake/guake/pull/572
I fixed the filechooser configuration
The custom command list is now inside a dedicated menuitem
The custom command json configuration file now must be edited according your example and supports submenus and descriptions.

As always I look forward to know what you think.

Thanks


Stibbons:
please run ./validate.sh, there is a pep8 error somewhere. I'll merge it soon after :)

Ozzyboshi:
I fixed the pep8 error (I hope).
I think now you can merge however my autopep8 reports errors on ./install-lib.py that I have not modified.
The reason is probably the comment style (a series of # instead of a single one).

Stibbons:
merged!

Issue #538, opened 2015-05-13
Guake does not save window height


pensnarik:
When I change window height, hide terminal and restore it - I see the deafault window height again (50%). Version 0.7.0, but I can reproduce this in version 0.6.0 too.
I thought that guake cannot save configs, but I could change a background image, for example, and it works.
OS: Arch Linux.

Stibbons:
Hi.
Of course guake change config, they are in gconf (/app/guake/general/window_height). I can change it. So I think you are in a particular case where config cannot be saved... Did you have restarted after your installation?

pensnarik:
I have restarted system after installation, all files in ~/.gconf/guake is owned by me. But it still does not work. I can change window height running gconf-editor and setting parameter manually.

mouseratti:
Hello. I watch the same behaviour. I  change a window height with a mouse. Then I hide guake and show it again. Guake still has a previous window height. If I change it with preferences dialog, it has a new height
Guake: 0.8.0 
OS: 3.13.0-37-generic #64-Ubuntu
DE: Xfce

thegitfather:
Same here. Trying to run some commands to modify the appearance when i trigger a global shortcut. Changing font settings works though:

```bash
$ gconftool -s /apps/guake/style/font/style -t string "Monospace 16"
```

proowl:
+1

SimLemay:
I have this exact problem on Guake 3.5.0 - Ubuntu 16.0.4
I own everything in ~/.gconf/apps/guake
I can change the default height in Preferences => Main Window

PR #611, opened 2015-09-11
fixed issue #538

mouseratti:
Hello. I think I've fixed it. It works as expected on my station.


Stibbons:
please don't change line width. We use line width 100 char, as defined in pep8rc/pylintrc, which is pretty common nowdays

Stibbons:
merged

PR #612, opened 2015-09-11
Update INSTALL

selivan:
Instructions in INSTALL file are outdated. I have built guake 0.8.0 on Ubuntu 14.04 and added necessary steps.


Stibbons:
thanks. INSTALL is not really sync indeed

Issue #616, opened 2015-09-15
Big tab bar


tristan202:
All of a sudden my tab bar grew to double height, and I have no way of getting it back to normal size. I tried removing everything, and reinstall from scratch, but it's still doublee size.
![untitled](https://cloud.githubusercontent.com/assets/1045501/9887778/b63815f4-5bf1-11e5-9b96-9eb61bcfadc7.png)

Ozzyboshi:
try my new data/guake.glade file (https://github.com/Ozzyboshi/guake) , i built a new virtual machine with xfce on it and it seems to work, if yes i can submit a pull request to @Stibbons on this repository

Ozzyboshi:
i think this issue is a duplicate of #599

tristan202:
Resolved in my end.

Issue #599, opened 2015-08-19
New Tab icon not found and very big tab bar


Bubu:
See screenshot: 
![screenshot - 08192015 - 01 42 16 pm](https://cloud.githubusercontent.com/assets/105185/9355342/c820a1f2-4678-11e5-84d0-e5451a345f22.png)

I'm not sure but I guess the two problems are related.

System is Xubuntu 14.04 and I installed guake 0.8.0 from this ppa: https://launchpad.net/~webupd8team/+archive/ubuntu/unstable

Bubu:
Mmh, weird. Their is no reference to the add_tab.png in the current source files. Am I missing something?

Ozzyboshi:
If you look at old git commits you'll find this:

commit c6314aea554ed28780acd45ce4e2dfe8ffdb6c43 (2015-05-15 10:04:45)

In this commit the following lines were removed from the guake_app.py file 

-        ipath = pixmapfile('add_tab.png')
-        self.get_widget('image2').set_from_file(ipath)

I think this is why the icon is missing (I am experiencing same problem on my ubuntu 14.04)
The question is: why @Stibbons  did this? Maybe he wanted to set the image from glade.
Waiting from clarifications from @Stibbons

Stibbons:
indeed, this can be questionnable... I am rewriting everything from the bottom up so I hope to handle it properly in 1.0.0...

Bubu:
Ok, I re-added thse two lines to my local install and the icon does indeed appear again. But unfortunately the huge tab bar remains. Any idea where this is coming from?

Ozzyboshi:
Maybe the tab bar got bigger because a few months ago i added a scrolledwindow and a tabviewport in data/guake.glade.
When I did this I didn't notice visual changes on my ubuntu, can you try deleting this 2 components or using an old guake.glade file and see if something changes?

Bubu:
Yes, that seems indeed to be the case. Using the guake.glade from before this commit 652cd8565b1d52ef1fbce0ece4121ad24ce8e255 the tab bar is back to the old height.

You said you didn't notice any visual change in your config. Can I provide any more debug info why this problem happens on my setup?

Ozzyboshi:
In the next days I will spin up a new virtual machine with XUbuntu 14.04 installed on it and see if i can do something for the glade file (but i really doubt it).
[...]

PR #621, opened 2015-09-22
Fixed guake.glade for xfce and re-added tab png icon on tab bar

Ozzyboshi:
This PL request is should solve this issue:

New Tab icon not found and very big tab bar #599 and issue #616 , the latter was a duplicate.


Stibbons:
merged

PR #629, opened 2015-10-06
Update Russian translation (file "ru.po" that was present in repo, i can't update it)

vantu5z:
Update of "po" files by command "make update-po", not working for me. Or I do that not correctly.


Stibbons:
you have to be in po folder. Thanks for the update

PR #630, opened 2015-10-06
Update ru.po and translation

vantu5z:
You don't say that I need to run ./autogen.sh to make the Makefile in po folder.
After this I can update ru.po.


Stibbons:
Indeed :) There is also the .dev.sh script that can help

PR #631, opened 2015-10-06
Updated german translation

airfunker:
I went through the PO file for the german translation and corrected some strings.


Stibbons:
thanks a lot!

airfunker:
You're welcome :)

PR #650, opened 2015-10-25
Closes #262 and closes #350 Feature keyboard shorcut hide on lose focus

### LINKED ISSUE #262: [Feature request] Add "Hide on lose focus" toggle keyboard shortcut
I normally use Guake with "hide on lose focus" ON.

Occasionally though, I will perform some action (say in my browser) for which I need to observe the live debug in the console (Guake).

In such instance it would be very handy to have a keyboard shortcut to toggle the behaviour, rather than having to right-click on preference, open the tab on keyboard shortcuts, uncheck the "hide on lose focus" box [and the reverse when I am done].

### LINKED ISSUE #350: Keyboard shortcut to toggle hide on lose focus
It would be nice to have the option to add a keyboard shortcut to toggle hide on lose focus

thardev:


Stibbons:
thanks!

PR #654, opened 2015-10-28
Fix #651 and hide on lose focus small changes

### LINKED ISSUE #651: Endless transparency keyboard shortcut toggles
Hello! I've noticed that when you use the keyboard shorcut to increase/decrease transparency if you keep pushing the key combination, the transparency goes to the higher level but, as I see in the code the transparency function that controls the level, doesn't have a limit, so you can increase it endless and the result is that when you want to decrease it doesn't works as expected...

Code: https://github.com/Guake/guake/blob/589131ac2383dd871d20dd34a3501717b09039ad/src/guake/guake_app.py#L1117-L1129

PS: It's a bit hard to explain this issue because I cannot show it with screenshots, GIF's, etc :tired_face:

thardev:


Stibbons:
thanks!

PR #655, opened 2015-11-02
Closes #47 new feature max tab name length

### LINKED ISSUE #47: Allow setting a maximum size for the tab vte title
Allow setting a maximum size for the tab vte title  http://guake.org/ticket/504
Reported by: 	clopez 	Owned by: 	somebody
Priority: 	Fix available 	Milestone: 	0.5.0
Component: 	guake 	Version: 	
Keywords: 	patch, tab, title, vte 	Cc: 	clopez@…
Description

Since #164 guake can display the tab title.

The problem is that as soon as you start digging into the files system, many times the title of each tab grows until you end with all the horizontal space filled by just 4 or 5 tabs.

This is very annoying.

The attached patch adds a new configuration variable that allows to put a limit on how much a table title can grow.

[PATCH] Allow setting a maximum size for the tab titles:

 * Change the value with:
   gconftool-2 --type integer --set /apps/guake/general/max_vte_title_size N

    * if N = 0 disables the maximum size (unlimited)

    * Otherwise cuts the tab titles at the Nth char.

thardev:
Hello! :smiley: I have implemented the requested feature that limit the tab names to the Nth character.
I have implemented this way:
* Added a Spin button in preferences to set the value, in the General --> Main Window section.
* Values allowed are from 0 to 100. 
* 0 equals unlimited length.
* Every time when the value is changed, it triggers the function that sets the new name of the tabs.
* When a tab it's renamed, stores the name in a new *attr* of the tab and shows the name cut at the Nth char.

That's all! Don't hesitate to ask about any changes :wink:

### DIFF
--- a/data/guake.schemas
+++ b/data/guake.schemas
@@ -424,6 +424,18 @@
                 <long>Any terminal program can set the terminal's title via special escape sequences. Some shells (e.g. bash) display user's prompt there (though you can adjust that to any arbitrary text in bashrc or similar) and update it automatically as the prompt changes. So you can have easy automatically updating meaningful tab titles with this option turned on.</long>
             </locale>
         </schema>
+        
+        <schema>
+            <key>/schemas/apps/guake/general/max_tab_name_length</key>
+            <applyto>/apps/guake/general/max_tab_name_length</applyto>
+            <owner>guake</owner>
+            <type>int</type>
+            <default>0</default>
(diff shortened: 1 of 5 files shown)


Stibbons:
thanks !

thardev:
Thank you!
Nice! welcome back @Stibbons hope you enjoyed! :wink:

PR #709, opened 2016-02-14
correction for Fedora install from F23.

gbonnema:
From Fedora 23 the repositories are accessible through "dnf" in stead of "yum". 1 small correction to a readme to show the diff.


pypingou:
Looks good and is valid, merging :)

Stibbons:
+1

PR #712, opened 2016-02-20
Now use spawn_sync from newer versions of VTE

Jiehong:
I was giving a try at the gtk3 version of guake (not on branch gtk3, how odd), and I couldn't because of newer version of VTE renamed `fork_command_full`.


Stibbons:
Yes it depends on the version. I am working on a little abstraction to this function rename.

Issue #713, opened 2016-02-27
Feature request: Paste on right-click


samuelhug:
I would like to add an option to enable paste on right-click.
The user could still open the context menu by doing ctrl+right-click.

PR #714, opened 2016-02-27
Add option for paste on right-click #713

samuelhug:
Adds an option in the preferences to enable pasting on right-click. The context menu can still be accessed via a ctrl+right-click.

### DIFF
--- a/data/prefs.glade
+++ b/data/prefs.glade
@@ -144,7 +144,7 @@
                                   <widget class="GtkTable" id="table5">
                                     <property name="visible">True</property>
                                     <property name="can_focus">False</property>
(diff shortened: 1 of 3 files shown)


Stibbons:
I don't see the point of this feature. Middle click is more classic to use... What do the others think about such feature?

samuelhug:
My personal use case is a laptop with a touchpad and no middle mouse button.

I understand if its not a feature other people want. Just thought I'd share
in case other people wanted it.

pypingou:
There is often a way to simulate a middle click action with either pressing the left and right button at the same time, or using something like 1 finger tap = left click, 2 fingers tap = right click, 3 fingers tap = middle click

PR #716, opened 2016-02-29
Remove mention of global hotkeys, was replaced with keybinder

ElijahLynn:
It appears that global hotkeys was replaced with keybinder in https://github.com/Guake/guake/issues/510.


ElijahLynn:
Ahh, looks like this https://github.com/engla/keybinder is actually C still.

PR #732, opened 2016-04-07
Fixed german translation, updated prefs-frontend

EifX:
Moved the combo-box a little bit to the right side for larger translations


Stibbons:
thanks !

PR #734, opened 2016-04-14
Fixed german translation, modified prefs a little bit

EifX:
- Fixed translation, added some strings
- Modified prefs for displaying larger lang-strings


Stibbons:
thanks

PR #737, opened 2016-04-22
Use default value for Unity icon-size if it is not present in dconf

ashkop:
If Unity icon size was never changed, dconf property `/org/compiz/profiles/unity/plugins/unityshell/icon-size` is absent. This causes `int()` call to throw an exception and `found` will be `False` even though it should be `True`.


Stibbons:
good

PR #744, opened 2016-05-09
Update pt_BR.po

fit087:
"_Reset Compatibility Options to Defaults" can translated better as "_Restaurar as opções de compatibilidade padrão"


Stibbons:
thanks!

PR #761, opened 2016-06-25
Shorter "Open Link" menu

thuandt:
Long URL cause right click menu is very ugly.
We should make URL shorter, only need display 25 first characters.

#### Now
Original
#### After
Patched

### DIFF
--- a/src/guake/guake_app.py
+++ b/src/guake/guake_app.py
@@ -646,7 +646,7 @@ class Guake(SimpleGladeApp):
         link = self.getCurrentTerminalLinkUnderCursor()
         if link:
             self.get_widget('context_browse_on_web').set_visible(True)
-            self.get_widget('context_browse_on_web').set_label(_("Open Link: {}".format(link)))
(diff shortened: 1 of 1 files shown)


src/guake/guake_app.py
```diff
@@ -646,7 +646,7 @@ def show_context_menu(self, terminal, event):
         link = self.getCurrentTerminalLinkUnderCursor()
         if link:
             self.get_widget('context_browse_on_web').set_visible(True)
-            self.get_widget('context_browse_on_web').set_label(_("Open Link: {}".format(link)))
+            self.get_widget('context_browse_on_web').set_label(_("Open Link: '{}...'".format(link[:25])))
```
Stibbons: add a if condition
```
if len(link) >= 28:
   self.get_widget('context_browse_on_web').set_label(_("Open Link: '{}...'".format(link[:25])))
else:
  self.get_widget('context_browse_on_web').set_label(_("Open Link: {}".format(link)))
```

Stibbons:
great! Thanx!

PR #773, opened 2016-07-18
fix correct screen

### LINKED ISSUE #770: Height issue moves Guake to non-primary screen
When Guake is set at 100% height it forces the terminal to the secondary screen. Probably because it exceeds the viewport size due to the top bar height.

Using Gnome 3.20, Ubuntu 16.04.

koter84:
i made a change to maximize() the window instead of resize and move when width and height are set to 100%

this fixes #770 and always gets the window on the correct monitor


Phiwa:
:+1:

Stibbons:
thanks !

PR #774, opened 2016-07-18
fix fullscreen follow mouse

koter84:
i noticed that when in full-screen-mode the window doesn't follow the mouse. i don't know if this is intended behavior or a bug. seems to me that if you want to fix Guake to a specific monitor you can set that up in the config.

to change the behavior i had to change something in a part which had a comment about a work-around for an issue in fluxbox, i don't have fluxbox so i wasn't able to test any regressions there.

please let me know if this is an OK way to make the full-screen-mode follow the mouse

### DIFF
--- a/src/guake/guake_app.py
+++ b/src/guake/guake_app.py
@@ -784,8 +784,7 @@ class Guake(SimpleGladeApp):
         self.hidden = False
 
         # setting window in all desktops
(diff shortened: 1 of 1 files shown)


Phiwa:
:+1:

Stibbons:
Look good to me

Issue #723, opened 2016-03-22
Always need to drag the handle on tabbar to resize the window (even if the tabbar is not on the side of the window border)


hengyang-zhao:
How to replicate:

Check "Place tabs on top" in preferences
Try resizing the guake window by dragging

aichingm:
The resize handle has been removed from guake -> closing.

Issue #382, opened 2014-10-12
Resizing issues


bergfried:
There are several glitches when you try to resize the window:
- If you resize the window using your window manager's capabilities, guake ignores the actual window size when you try to resize it using guake's shortcuts afterwards. It would be nice to see the window's dimensions being well-defined. That is, every time I use the window manager to resize (or maximize/restore) the window guake should automatically resize the window so that the resulting dimensions are close to the requested dimensions as well as compatible to guake's own constraints.
- If "Bottom align window instead of top align" is enabled the divider (i.e. the small bar you use to resize the window using your mouse) behaves as if the window is still being aligned to the top which is really strange.
- Also, if "Bottom align window instead of top align" is enabled said divider should appear at the top of the window instead of at the bottom.

gammy:
I have a more serious resizing issue - maybe this isn't the right place to report it. A video seems a more apt way to demonstrate the issue.

In short, resizing is unusable on my setup, where i have three screens in this fashion: 
vertically oriented, horizontally oriented, vertically oriented (`[] [     ] []`), with guake being in the center.

https://www.youtube.com/watch?v=cqp4mh9yt9E

PR #775, opened 2016-07-18
Fixing 'resizer' position in tabs-on-top mode and resizing while bottom-aligned

RedFlames:
Hello,
I noticed that the little resizer handle will not be at the bottom when you switch the tab-bar to the top, and also that the window will not behave correctly when resizing it while bottom-aligned. I wrote some quick fixes, although they could be better.
I still get some flickering on bottom-align resize because something tries to adjust for my taskbar... and the code to always push the resizer element to the bottom might be better of in a small helper function on its own instead of in set_tab_position.
(With a little more polish, maybe) this would fix issue #723 and 2nd/3rd point of issue #382.

### DIFF
--- a/src/guake/gconfhandler.py
+++ b/src/guake/gconfhandler.py
@@ -145,6 +145,7 @@ class GConfHandler(object):
         be called and will call the move function in guake.
         """
         self.guake.set_final_window_rect()
+        self.guake.set_tab_position()
 
     def size_changed(self, client, connection_id, entry, data):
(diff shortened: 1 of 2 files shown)


Stibbons:
thanks !

brainiac05:
Somewhat related: Is there a way to disable the resize handle?

brainiac05:
Somewhat related: Is there a way to disable the resize handle?

PR #777, opened 2016-07-21
Fix #636 by killing preview terminal

### LINKED ISSUE #636: guake creates zombies on gnome 3.18
screenshot from 2015-10-13 08-46-40

coderstephen:
Fixes #636.

The issue lies in the terminal preview embedded in the preferences window. Even when the terminal was destroyed, in certain situations the shell process did not. This was especially true when the shell process was something that expected to be long-lived, like tmux.

This fixes the issue by ensuring that the shell process is killed whenever the preferences window is destroyed.

### DIFF
--- a/src/guake/guake_notebook.py
+++ b/src/guake/guake_notebook.py
@@ -5,11 +5,8 @@ from __future__ import print_function
 import logging
 import os
 import posix
-import signal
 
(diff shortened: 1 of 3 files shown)


Stibbons:
thanks!

PR #786, opened 2016-08-04
Fix the bottom of tab buttons being cut off in Gnome

### LINKED ISSUE #666: Tab selector cut off
The tab selector bar has is cut off on the bottom, as shown in this screenshot. I am running Guake 0.8.3 and Gnome 3.18.3 on Arch Linux.

screenshot from 2015-11-18 11-27-09

lopsided98:
A simple patch that fixes #666 by removing the explicit height request from `tabs-viewport`.

This is the result of this change for me:
* Without scrollbar: image
* With scrollbar:
image

I'm not sure if this will have any unintended side effects on other desktops.

### DIFF
--- a/data/guake.glade
+++ b/data/guake.glade
@@ -427,7 +427,6 @@
                 <property name="vscrollbar_policy">never</property>
                 <child>
                   <widget class="GtkViewport" id="tabs-viewport">
-                    <property name="height_request">10</property>
                     <property name="visible">True</property>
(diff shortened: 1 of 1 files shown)


Stibbons:
look good

versusvoid:
Tabs now looks awkwardly big and empty. A nice solution would set height_request on scrollbar appearance and reset on disappearance.

Issue #653, opened 2015-10-27
Scrollbar hide tabs names


thardev:
Hello! I'm going to attach a screenshot because it's self-explanatory :+1: 
![screenshot from 2015-10-27 21 09 44](https://cloud.githubusercontent.com/assets/8918134/10771085/28298cec-7cef-11e5-8d6a-f730818c22cb.png)

As you can see the horizontal scrollbar appears and hide the tabs names.
I use Debian Jessie 8.2 and Gnome Shell 3.14.4

MAD-GooZe:
Happens also in Ubuntu 15.10 with Gnome Shell 3.18.1

tiltkoko:
is there a method so the tab names don't get too long?

thardev:
Hello @tiltkoko 
My pending pull request it's going to implement just that functionality, limit the tab names as a preference, on the other hand I'm working on this issue and it's nearly to be solved but I'm waiting to my actual pull request be merged :+1:

Ozzyboshi:
Hello @thardev  , did your pl solve the scrollbar issue or just the tab names limit? I don't see any changes in the guake.glade file. I reproduced this bug on a new virtual machine and I am willing to fix it.

thardev:
Hello @Ozzyboshi 
In my last pull request, I've solved just the tab names, but the scrollbar issue I'm working on it but I'm having some troubles getting the first tab size (I'm solving this programmatically, not just modifying the glade file. Calculating the size of the tabs and getting the width of the screen) I think I'll work on it and it will be ready in short time :smiley:

Aminelahlou:
I have the same issue

r-barnes:
Affects me with 0.8.4 on Lubuntu 16.04.

thrantir:
Affects me too
ubuntu 16.04
guake 0.8.4-1

Mirkbot:
Same here on Fedora 24 with Cinnamon and Guake 0.8.5

versusvoid:
Solution (alongside with #666):
https://github.com/versusvoid/guake/commit/444b151fc5c7327645e575ff4ce6786cf61fbf65

Got a little bit confused with two similarly named issues, sorry.

Stibbons:
Should have been fixed

PR #798, opened 2016-08-17
Abbreviate tab names

versusvoid:
Better fixes for #653 and #666 (preserving tab height)
Option to abbreviate directory paths in tab names when necessary.
Small preference dialog fixes.


Stibbons:
thanks i ll have a look

PR #801, opened 2016-08-18
allow abbreviation independently of max_tab_name_length, fix behaviour

versusvoid:
I want abbreviate tab names when necessary, but don't want to cut them
Additionally, fix initial abbreviation and possible unwanted abbreviations before current directory in vte title.


Stibbons:
ok, looks good

PR #811, opened 2016-09-17
Notifications/close other tabs/open link

berdario:
These are a few small changes that I and a friend of mine worked on 5 years ago. Unfortunately at the time Guake was in a bit of a limbo... we started development on Gitorious, but we weren't sure where to submit the PR after that, and then we simply forgot about it.

Now that development is done here on github, I noticed these changes while going through a bunch of old repositories of mine...

This doesn't merge cleanly, and it's possible that similar changes might have been implemented already... but I guess that this PR could still be useful to gauge interest, and if that's the case, I could try to clean up a bit before resubmitting

### DIFF
--- a/data/guake.glade
+++ b/data/guake.glade
@@ -210,6 +210,21 @@
       </widget>
     </child>
     <child>
+      <widget class="GtkImageMenuItem" id="context_open_link">
+        <property name="label" translatable="yes">Open link</property>
+        <property name="visible">True</property>
+        <property name="use_stock">False</property>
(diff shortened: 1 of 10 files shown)


data/guake.glade
```diff
@@ -210,6 +210,21 @@
       </widget>
     </child>
     <child>
+      <widget class="GtkImageMenuItem" id="context_open_link">
+        <property name="label" translatable="yes">Open link</property>
```
Stibbons: there is already this link

PR #816, opened 2016-09-26
Added support for selected text in custom commands

Ozzyboshi:
No words from @skraelings (see issue #564) anyway this feature could be useful, asking for a PL.
Here's the description.

Now it's possible to pass the in terminal selected text to custom commands, to do that just put the number 1 in the command array withing the custom command file.
For example the following json entry launches the command "stat #filename#" and filename is the terminal selection:

...
{
        "description": "File stat",
        "cmd": [
                    "stat",
                    1
                ]
},
...

Note that the '1' flag must be and integer and not a string, so avoid enclosing in double quotes.
Other numbers could be used for future features.
If there is not a selection the command "stat" is issued with no parameters.

### DIFF
--- a/data/custom_commands_help.glade
+++ b/data/custom_commands_help.glade
@@ -1,0 +1,92 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<interface>
+  <requires lib="gtk+" version="2.24"/>
+  <!-- interface-naming-policy project-wide -->
+  <object class="GtkAssistant" id="assistant1">
+    <property name="can_focus">False</property>
+    <property name="border_width">12</property>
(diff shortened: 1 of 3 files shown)


Stibbons:
Can you update the README to document this feature?

Ozzyboshi:
I think the NEWS file would be a better place to document this feature, the root README.rst mostly describes how to compile/run Guake and how to contribute.
What do you think @Stibbons

Stibbons:
Best place is in a tooltip or in a label in the preference windows

Stibbons:
you need to rebase (don't know why)

Issue #32, opened 2013-04-14
Workspace-specific guake


kikecalpe:
Workspace-specific guake http://guake.org/ticket/543
Reported by:    PietroB     Owned by:   somebody
Priority:   major   Milestone:  0.5.0
Component:  guake   Version:    0.4.4
Keywords:       Cc:     me@…
Description

I just discovered guake, and I felt in love with it: it's nice and fast (and well documented Python code too!).

But I have another love: workspaces. And it clearly makes no sense that the same shell which I just used to batch rename some songs in my "Free time" workspace comes out in my "University" workspace. I want another one.

Please put an option so that guake can be ran as workspace-specific.

This requires two steps: - guake should not run on all workspaces - when re-run, it should trigger via dbus the existing instance only if it's in the same workspace

Actually, I'm willing to try to produce a patch. But I feel very stupid: looking at guake's code (and glade file!), I wasn't able to understand _how_ it sets its presence on all workscreens.

An alternative, maybe nicer, is that the instance is single, but guake shows different terminals depending on the workspace. Nicer, but probably harder (at least, I would have to dig more in guakes' internals).

cikaRadule:
Hi, guys what has happened with this feature request?
Seems it's dead in the water for more than two years now. :/
I could try to pick it up, but can't seem to find any current work on this issue...

Stibbons:
no found time to work on it. Would be part of next rework of guake with gtk3, but it does not have advanced that much sadly :(

toobaz:
I have been using my version modified with workspaces until recently, but it diverged since long, not sure when I'll find the time to rebase. Anyway, it still had some bugs (apart from the fact that the functionality could not be switched on/off).

cikaRadule:
Is there any chance I can take a look at your work, at least to try to make it run on my machine?
And if possible to try to see if it can be prepared for some future releases...

Stibbons:
work has not much advanced... there are two attempt on (guake3 and gtk3 branches). But you can rebase on master and submit a pull request on the current version

toobaz:
@cikaRadule see #820

cikaRadule:
@toobaz thanks, I'll take a look and hopefully manage to make some progress. I'll leave status update here....

AlexWayfer:
Some updates? :sweat:

candoz:
I want to believe

Stibbons:
I unplan this feature, no time to work on it. I am open to pull request for some that wants to dig it :)

[...]

PR #820, opened 2016-10-01
Multiws

toobaz:
Here is an old attempt at workspace-specific guake. It is far from ready, I'm opening the PR just in case it is useful to someone else (see  #32).

Doesn't even merge at the moment, but I'll try to solve this soon.

More important is to understand whether the strategy I adopted, which was the simplest possible, was also the best one. What I do is to simply keep a single instance of the terminal, but hide the tabs (which are all in a same notebook container) depending on which workspace is shown. Maybe a better approach would be to have more conceptually separated instances (maybe even processes!), which could also make it easier (just guessing) e.g. to "open a terminal here" in nautilus.

Apart from this, the current implementation has at least a couple of limitations:
- the functionality cannot be switched on/off
- I remember some occasional unexplained bug when closing a tab
  ... but apart from that, I really liked it (and this is why I plan to at least rebase to make it usable again).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -75,6 +75,9 @@ Dependencies
  * python-notify
  * python-vte
  * python-xdg
+ * python-wnck
+ * python-appindicator (ubuntu)
+ * notify-osd (ubuntu)
  * libutempter
 
 To build guake, you will need the following packages too:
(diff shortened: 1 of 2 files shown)


Stibbons:
can you rebase?

toobaz:
Rebased, but I didn't have time to test at all.

Issue #695, opened 2016-01-28
Configuration via escape sequence


milgner:
Is there a way to configure Guake using escape sequences?

I currently use Terminology as my terminal emulator and when I ssh in to important server systems, I set a corresponding background image via escape sequence so I can see at a glance when such a session is still open.

milgner:
I did some digging into the [python-vte](https://github.com/linuxdeepin/python-vte/) code and at first glance it looks like all escape sequences are hardcoded in [vteseq-n.c](https://github.com/linuxdeepin/python-vte/blob/master/src/vteseq-n.c) without any possibility to register additional ones. I'll do some additional investigation and try to come up with an idea...

Stibbons:
seems a bad idea to hack in the escape characters. we can react to other events to replace the background image, but indeed, it is not possible for the moment to customize image per tab

ulukyn:
I have made some changes to be able to set a custom image per tab using --bgimg parameters.
No idea if my changes are right and completes. But works nice for me using command like

guake -s 1 --bgimg="path_to_image/image.jpg" -r PrettyTab


Attached patch:
[image_per_tab.patch.txt](https://github.com/Guake/guake/files/506716/image_per_tab.patch.txt)

aichingm:
Setting background images is not supported anymore. This has changed in the vte 2.90 or something like this.

Issue #369, opened 2014-09-05
Feature request : Multi-background images


polygbrl:
Priority : Low

Using guake a lot, I found myself tired of having the same background-color / image each time I make guake scroll down. While I'm not a customization fan, I must say that it could be nice to select multiple images as backgrounds and make it change from one image to an other each time we make Guake pop.

Could be a fancy idea.

ulukyn:
I have made changes to be able to set the background image using the command line
Here the issue : https://github.com/Guake/guake/issues/695

So, with the new --bgimg, It's possible to bind the F12 key (or another) to a script who will change the background before show Guake. Something like :

`#!/bin/sh`
`# The path of all images to set randomly`
`IMG_PATH="/xxx/xxx"`
`TAB=$(guake -g)`
`SELECTED_IMG=$(ls $IMG_PATH |sort -R |tail -1)`
`echo "$SELECTED_IMG"`
`guake -s $TAB --bgimg="$IMG_PATH/$SELECTED_IMG" --show`

Stibbons:
Backgroup image no more supported

cloudsemail:
Dang.... the background image was one of the coolest things about guake... no chance it'll come back?

gsemet:
I’ll be glad to reintroduce it once vte support it again

PR #837, opened 2016-11-06
Added command line option to setup the background image of the tab

ulukyn:
To answer to feature request : Multi-background images #369 and Configuration via escape sequence #695


Stibbons:
look good

PR #851, opened 2016-11-24
[feature] Close a tab with the middle button of the mouse

mnajle:
Closing a tab with the wheel button of the mouse is just a feature that is very useful for me because I'm really used to do it in all the applications I deal with.


pylintrc
```diff
@@ -77,6 +77,8 @@ disable=
   C0322,
   C0323,
   C0324,
+  C0411,
```
Stibbons: i would like to keep these

Issue #874, opened 2016-12-13
add tests to setup.py in guake3


mouseratti:
I'm planning to enable pytest for **python setup.py test** command

Stibbons:
can you do this on the "guake3" branch ?

mouseratti:
Sure

PR #878, opened 2016-12-19
#874 pytest added as test framework

mouseratti:


mouseratti:
@Stibbons could you review it, please?

setup.py
```diff
@@ -7,6 +7,7 @@
 from setuptools import setup
 
 setup(
-    setup_requires=['pbr'],
+    setup_requires=['pbr', 'pytest-runner', 'wheel'],
```
Stibbons: why don't you use `requirements-dev.txt`?

setup.py
```diff
@@ -7,6 +7,7 @@
...
+    tests_require=['pytest', ],
```
Stibbons: isn't there a way to insert it into setup.cfg and keep setup.py as minimal as possible?

Also, useless comma. `tests_require=['pytest']` works.

setup.cfg
```diff
@@ -24,7 +24,6 @@ keywords =
 [files]
 packages =
     guake
-
```
Stibbons: please keep this line

mouseratti:
current PR is closed, see #880

PR #885, opened 2017-01-07
Homogeneous tabs

tombriden:
Slightly different way of handling the tabs where they share the full screen width. So a single tab is full width, 2 are half width each etc. Basically how things like gnome-terminal behave
I've also added a tooltip so the title can be read if it's too long to read in the tab


eli-schwartz:
Could this be made optional? I kind of really don't like the way this looks at all and would vastly prefer the old way...

Issue #565, opened 2015-06-08
Guake not taking the entire width of the screen


kookerus:
For a while, Guake has left some ten or so pixels of empty space on the right side of the screen (screenshot below because I suck at describing things). I have tried aligning center, left, and right, and nothing fixes it. I also have a screenshot of my preferences. Using the most current version of master as of the time of writing.
![gap](https://cloud.githubusercontent.com/assets/8160403/8029161/b3bd8984-0d87-11e5-9d71-489e2d05b2ce.png)
![preferences](https://cloud.githubusercontent.com/assets/8160403/8029160/b3bd3a42-0d87-11e5-8cc8-005d62649f0c.png)

pypingou:
Could it be the space for the scrollbar?

Stibbons:
can you describe your system, version, gnome version, if you have 1 or 2 screens?

kookerus:
@pypingou it isn't, the scrollbar is contained in the terminal

@Stibbons Running a clean install of Ubuntu 14.04.2, on Gnome flashback (I believe it's version 3.9.90), using Metacity (compositing enabled), and I have 1 screen (1920x1080).

Stibbons:
Kinda reproduced. I've installed gnome-fallback on ubuntu 15.04 and have the following issues:
- hotkey only work when mouse is hover the main bar
- full height is not taken into account

kookerus:
That's odd. It doesn't bother me, I just figured I'd report it to be safe =P

Stibbons:
what do you mean by "compositing enabled" ?

Stibbons:
what do you mean by "compositing enabled" ?

kookerus:
You can make metacity look/act like Compiz by enabling compositing in the dconf-editor.
If the dconf key "org>gnome>metacity>compositing-manager" is set to true, it enables true transparency, as well as some other features of compiz.

kookerus:
Disabling compositing does not fix the issue.

Stibbons:
ok i 'll check what i can do with what i have reproduced...

Vizzard:
I have this on Ubuntu 15.04 (Unity) with Guake 0.7.2 (laptop with one screen).
Changing guake-0.7.2/src/guake/guake_app.py line 871:
window_rect.width = window_rect.width - unity_dock
like this by removing "- unity_dock":
window_rect.width = window_rect.width
and running 'dev.sh --reinstall' and 'dev.sh --install' fixed issue for me.
[...]

Issue #522, opened 2015-05-05
Gap on the right in Ubuntu 15.04


Suor:
I just updated to Ubuntu 15.04, Guake 0.5 and now I have a gap on the right, although guake width is set to 100%. If I set align right the gap persists, so it's not the incorrect width bug.
![guake-gap](https://cloud.githubusercontent.com/assets/284103/7471142/a361faba-f34d-11e4-8a8f-d1d6be5b1ae5.png)

Suor:
Updated to 0.7, and this is fixed there. Closing, sorry.

Vizzard:
I don't believe it's fixed. I have this on Ubuntu 15.04 with Guake 0.7.2.
Changing guake-0.7.2/src/guake/guake_app.py line 871:
window_rect.width = window_rect.width - unity_dock
like this by removing "- unity_dock":
window_rect.width = window_rect.width
and running 'dev.sh --reinstall' and 'dev.sh --install' fixed issue for me.
Looks like 'window_rect.width' already has correct value because guake is full width when launcher is hidden and when launcher is visible it also has correct size (100% screen width minus launcher width). It's working even with different launcher size set (both smaller and larger than default).

Davidy22:
Fixes merged and original submitter marked resolved, closing.

Issue #422, opened 2014-11-27
Guake multi-monitor placement / unity dock width subtraction in monitor without dock


xfenix:
Hi, first of all, thank you for your great terminal!

![screenshot from 2014-11-27 19 39 01](https://cloud.githubusercontent.com/assets/1441415/5219808/21a579e6-766d-11e4-8657-105f1de22b06.png)

Probably, this is another problem with multi-monitor case.
I am place dock only for second, right monitor, and disable it for first, left (i attached a screenshot).
But, how you can see, guake anyway subtract 64 pixels from width.
I can "fix" this if replace `if unity_hide != 1` with `if False:` (and file guake_app.py probably, not guake.py). But, this is dirty, stupid, works only for me and only for this case.
Probably, there can be simpler solution (or im miss something). Or, if not, than, will be so cool, if you can add some option to avoid this problem in future releases. 
Thanks!

"Fixed" problem:
![screenshot from 2014-11-27 20 14 57](https://cloud.githubusercontent.com/assets/1441415/5220142/1888140e-7672-11e4-965f-ada2ea983b08.png)

Stibbons:
is it still applicable?

cryptobanana:
Greetings. I am experiencing a similar problem (ovserved both in verion 0.5.0 and version 0.7.0) where the width of the guake window does not extend across the entire screen. In my situation, however I have only one monitor in use. As illustrated in the screenshot, observe the gap between the edge of the guake window and the right-hand side of the screen.

![image](https://cloud.githubusercontent.com/assets/1322972/7554087/4a260804-f719-11e4-8659-271489fd7d4e.png)

P.S. Thanks for Guake :-)

Vizzard:
Changing guake-0.7.2/src/guake/guake_app.py line 871:
window_rect.width = window_rect.width - unity_dock
like this by removing "- unity_dock":
window_rect.width = window_rect.width
and running 'dev.sh --reinstall' and 'dev.sh --install' fixed issue for me.
Looks like 'window_rect.width' already has correct value because guake is full width when launcher is hidden and when launcher is visible it also has correct size (100% screen width minus launcher width). It's working even with different launcher size set (both smaller and larger than default).

tchaffee:
[...]

PR #893, opened 2017-01-17
Fix window width with Unity

brandoncurtis:
Addresses #422, #522, #565 when Unity dock is used on the bottom of the screen by adding a check for the dock's position; modify Guake's width to account for the Unity dock's width **iff** the dock is positioned "Left" or "Right".

### DIFF
--- a/src/guake/guake_app.py
+++ b/src/guake/guake_app.py
@@ -1008,10 +1008,12 @@ class Guake(SimpleGladeApp):
         if self.is_using_unity():
 
             # For Ubuntu 12.10 and above, try to use dconf:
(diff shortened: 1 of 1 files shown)


Stibbons:
look good

Stibbons:
Merged !

tombriden:
I think validate.sh was run on an older base as that commit has reverted the last few commits and guake no longer runs.

brandoncurtis:
@tombriden, correct, I messed up and forgot I checked out v0.8.8.  Opened a PR to revert, then I'll redo.

PR #929, opened 2017-04-05
Add quick open possibility for lines containing `/home/`

langpavel:
image

### DIFF
--- a/src/guake/terminal.py
+++ b/src/guake/terminal.py
@@ -72,7 +72,10 @@ QUICK_OPEN_MATCHERS = [
      r"^\s\sFile\s\"(.*)\",\sline\s([0-9]+)"),
     ("line starts by 'Filename:line' pattern (GCC/make). File path should exists.",
      r"^[a-zA-Z0-9\/\_\-\.\ ]+\.?[a-zA-Z0-9]+\:[0-9]+",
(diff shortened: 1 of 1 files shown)


Stibbons:
hum, ok for merge, even I would prefer if /anything that exist would be supported...

langpavel:
I also discovered that my code not work with paths containing `-` but I think my regex is correct. Can somebody help with this?

Also I have suggestion for character position capturing in regex, like `path/file:line:char` which is not possible now, but I'm not good in :snake: so I don't try implement it...

langpavel:
@Stibbons on POSIX/Linux systems `/home/` is enough. Matching everything which can be path will be probably too expensive on regex and syscalls. This is reasonable compromise

Stibbons:
Indeed. But every interesting files might not be on /home (/mnt or /media as well for example). Logs from /var, config from /etc, might also be interesting !

langpavel:
@Stibbons Right, we can add very tight distionary to regex, but really, very tight. So in POSIX systems it can be `home` as every developer works on local machine.
Interesting `/` directories: `home`, `root`, `etc`, `var`, `usr`, `opt`, `run`, `tmp`, `mnt`, `media`, `proc`

Issue #62, opened 2013-04-18
Hotkey for Grab Focus


kikecalpe:
Hotkey for Grab Focus http://guake.org/ticket/480
Reported by:    Nicolas     Owned by:   somebody
Priority:   minor   Milestone:  0.5.0
Component:  guake   Version:  
Keywords:       Cc:  
Description

Hi, this is a very great tool! I use it everywhere - at work, at home, in my dreams... kidding. ;-)

What I miss most is the possibility to grab the guake focus via hotkey, because guake is not in the alt+tab window switcher.

This would be a handy enchancement!

Regards, Nicolas

wranvaud:
Just having the option "Alt+TAB" in the preferences would be awesome

Stibbons:
I do'nt think this will be allowed by your window manager, since it is the standard hotkey for switching app...

wranvaud:
That's exactly what I would like, switch between apps, including Guake. I've done what smasty suggested to achieve this but a checkbox in preferences would be more user friendly ;)

MatthiasKauer:
@smasty Is that workaround still functional? 
I find a /usr/share/guake/guake.glade (instead of .glide), but it doesn't have that option on line 8. I changed the one I found further down but it didn't seem to help (after restart of guake only). Guake is nowhere to be found in Alt-tab list. This is guake 0.7.2 on debian testing with xfce 4.12.

smasty:
@MatthiasKauer Yes, it should still work.
In 0.7.2, the line is 378, and yes, the file is `/usr/share/guake/guake.glade`, there is a typo in my original comment.

Stibbons:
actually guake is made not to appear in alt tab and react only to our magic hotkey....

MatthiasKauer:
@Stibbons Yes, but if guake is not set to auto-hide I cannot get back into its window without using the mouse otherwise. Are there are other options?
@smasty Thank you!

silgon:
Got the same problem, whenever the gnome keyring is shown (let's say for a git pull or any ssh activity) guake gets out of focus and then it doesn't come back, so I have to click with the mouse or toggle hide/show twice. My nasty wayaround is a keybinding that does "xdotool click 1", because my mouse is usually in the position of the terminal. I'm using xubuntu 14.04, any news about this topic?

felipeespic:
No news yet? Having this feature would be really useful.

denis-gonzo:
I tried to deal with it, and here what I came to:
[...]

PR #938, opened 2017-04-22
Issue #62

denis-gonzo:
if window is displayed but not focussed, press on the show/hide hotkey would focus back the window instead of hiding it


Stibbons:
look good. thanks

PR #942, opened 2017-04-24
Quick Open with dashes in /home/ paths

denis-gonzo:
QUICK_OPEN_MATCHERS line containing /home/

### DIFF
--- a/src/guake/terminal.py
+++ b/src/guake/terminal.py
@@ -74,8 +74,8 @@ QUICK_OPEN_MATCHERS = [
      r"^[-a-zA-Z0-9\/\_\.\ ]+\.?[a-zA-Z0-9]+\:[0-9]+",
      r"^(.*)\:([0-9]+)"),
     ("line containing '/home/' absolute path",
-     r"(/home/[a-zA-Z0-9_\-\./]+)(:[0-9]+)?",
-     r"(/home/[a-zA-Z0-9_\-\./]+):?([0-9]+)?")
+     r"(/home/[-a-zA-Z0-9_\./]+)(:[0-9]+)?",
+     r"(/home/[-a-zA-Z0-9_\./]+):?([0-9]+)?")
 ]
 
 


src/guake/terminal.py
```diff
@@ -74,8 +74,8 @@
      r"^[-a-zA-Z0-9\/\_\.\ ]+\.?[a-zA-Z0-9]+\:[0-9]+",
      r"^(.*)\:([0-9]+)"),
     ("line containing '/home/' absolute path",
-     r"(/home/[a-zA-Z0-9_\-\./]+)(:[0-9]+)?",
-     r"(/home/[a-zA-Z0-9_\-\./]+):?([0-9]+)?")
+     r"(/home/[-a-zA-Z0-9_\./]+)(:[0-9]+)?",
```
Stibbons: do you know why `\-` does not work?

denis-gonzo: No, but I saw such behavior in different regexp for example in sed.

denis-gonzo: I checked in pure python re, `\-` works perfectly  
It seems that regexp doesn't work in vte.Terminal.
I don't know what happens after `vte.Terminal.match_add(match)`

denis-gonzo: This page http://www.regular-expressions.info/charclass.html :
> The closing bracket ], the caret ^ and the hyphen -  can be included by escaping them with a backslash, or by placing them in a position where they do not take on their special meaning. **The POSIX and GNU flavors are an exception. They treat backslashes in character classes as literal characters. So with these flavors, you can't escape anything in character classes.**

looks like python-vte uses POSIX regex

denis-gonzo: Also, quick open won't work with filenames with backslashes.
if you try to quick open this file:
`$ echo "text" > 'some\file.txt'`
`$ echo 'some\file.txt:1'`
command will be
`DEBUG    Opening file some\file.txt at line 1`
where \f if is an escaped char.

Stibbons: look like, thanks for this information !

PR #954, opened 2017-05-20
avoid UnicodeDecodeError when "guake -h"

zqb-all:
when I run "guake -h", I got 

> UnicodeDecodeError: 'utf8' codec can't decode bytes in position 1395-1396: invalid continuation byte

Finally I found that add a space between ` " ` and ` . ` can avoid this error.

### DIFF
--- a/src/guake/main.py
+++ b/src/guake/main.py
@@ -131,13 +131,13 @@ def main():
                       metavar='TITLE',
                       action='store', default='',
                       help=_('Rename the specified tab. Reset to default if TITLE is '
-                             'a single dash "-".'))
+                             'a single dash "-" .'))
 
     parser.add_option('-r', '--rename-current-tab', dest='rename_current_tab',
                       metavar='TITLE',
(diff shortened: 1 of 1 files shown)


Stibbons:
I am pretty sure this comes from your translation file. Adding a space just make the translation system find no translation for this string so stay with the default string.

Can you copy paste your environment variable (LC_LANG, ...) ?

zqb-all:
Before I do anything, I mean that when I clone the code first time, about a month ago, this error is there.
```
echo $LANG
zh_CN.UTF-8
echo $LANGUAGE
zh_CN:en_US:en

```
My Ubuntu 16.04 and Ubuntu 14.04 both have this error when I run "guake -h"

zqb-all:
You are right !

> Adding a space just make the translation system find no translation for this string so stay with the default string.

Before my translate, the translation file have problems, it contain chinese `"` . I will fix it later.  thx!

PR #975, opened 2017-07-08
[WIP] Added ability to modify shortcuts

mouseratti:
[WIP]  means "work in progress". You do not need to merge it at current state. I'll notify you when i make it ready to merge.


Stibbons:
don't you want to merge this even if it is not finished? maintaining a pull request is like having a branch, but less practical

PR #977, opened 2017-07-11
Add a configuration toggle to disable the refocusing #976

### LINKED ISSUE #976: [Regression] Key-Combination needs to be hit twice to close guake when it is not focused
Hay, I'm on Archlinux with guake 0.8.9 - 0.8.10.

1. Set the key combination to open to open guake to any thing (happens with every combination which I have tested)
2. Hit said combination which opens guake 
3.  Focus any other program or just click on the desktop
     * If this step is skip guake works just fine
4. Hit the combination to close guake: Nothing happens but `xev` shows that X fires both events. :-1: 
5. Hit the combination a second time and guake closes :+1: 

This is a real productivity bummer (at least for me).

If you need any more information please just ask. I would be really happy to help resolving this issue.

EDIT:
I did some more testing this happens with the versions 0.8.9 and 0.8.10, 0.8.8 works just fine!

aichingm:
This pr fixes #976 by adding a checkbox to the preference window

### DIFF
--- a/data/prefs.glade
+++ b/data/prefs.glade
@@ -730,6 +730,26 @@ Always</property>
 				  </child>
 
 				  <child>
+				    <widget class="GtkCheckButton" id="window_refocus">
+				      <property name="visible">True</property>
+				      <property name="can_focus">True</property>
+				      <property name="label" translatable="yes">Refocus if open</property>
+				      <property name="use_underline">True</property>
+				      <property name="relief">GTK_RELIEF_NORMAL</property>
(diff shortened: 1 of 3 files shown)


aichingm:
I think fixing the build would be out of the scope of this PR since a lot of files are involved which have not been touched by this commit.

aichingm:
Fixed formating. BTW what version of autopep8 are you using? running ./validate.sh changes about 15 files on my setup.

Stibbons:
I'll check the autopep8 and validation asap

PR #987, opened 2017-08-17
Update autogen.sh now gnome-common is obsolete

tombriden:
I've just replaced with the template from
https://wiki.gnome.org/Projects/GnomeCommon/Migration

and added the following line as suggested for intltoolize (which
itself is "almost obsolete"
`intltoolize --force --copy --automake || exit 1`


Stibbons:
might need more work on the autogen.sh script

tombriden:
Ok, this worked for me so must need some other changes for older versions of autoconf. Will get a VM going and take a look

PR #988, opened 2017-08-20
Move the startup script setting to the Hooks tab

aichingm:
Move the text input used for the startup script setting to the `Hooks` tab and label it with "On start:".
This change makes the `General` tab a little less cluttered and populates the quite empty `Hooks` tab. It also updates the German translation for the changed label.
guake-pref-startuphook


Stibbons:
thx

PR #1028, opened 2017-11-18
remove background image stuff (at least for now), not supported by vte anymore

aichingm:
I think it is a good idea to remove this feature as it is right now and reimplement it after the new version hits a stable state.


Stibbons:
Great :)

mlouielu:
WIP for the background image, reference the implement from https://github.com/gnunn1/tilix/commit/7ccb7f1c329ec7bb731f13d1f5ab5b157c2032f9


Guake!_978

PR #1047, opened 2017-12-29
Move the data in to the guake directroy to simpify the building of packages

aichingm:
Disclaimer: I'm not an build infrastructure nor package pro so please expect ruff edges. That said, this is the first version for which I was able to write a working PKGBUILD file for my archlinux machines. 

To get this to work I moved the data directory into the guake directory and changed the files setup.py and setup.cfg to include also non-python files so that the pixmaps and schemas will end up in the lib/python3.6/site-packages/guake/data directory.  If there is a better way to achieve this please let me know 👍 

I also added the old guake.desktop file which is not very well translated. Any ideas how to grab the translations out of the .po files and put them in to the .desktop file?  

Finally: I'm running only the new guake version on all my boxes 😄 

Edit: here is the PKGBUILD for any one interested. https://gist.github.com/aichingm/ed35ba3b136be4424b1ac947207dbca3 I will **not** load this up to the aur since I hope that it would be obsolete soon.


setup.py
```diff
@@ -15,4 +15,6 @@
     pass
 
 setuptools.setup(setup_requires=['pbr'],
-                 pbr=True)
+                 pbr=True,
+                 include_package_data=True
```
Stibbons: should be useless, pbr should already do it that, isn't it?

guake/globals.py
```diff
@@ -39,10 +39,13 @@ def is_run_from_git_workdir():
 VERSION = guake.__version__
 DATADIR = os.environ.get("GUAKE_DATA_DIR")
 
-IMAGE_DIR = 'data/pixmaps'
-GLADE_DIR = 'data'
-LOCALE_DIR = 'locale'
+SRC_DIR = os.path.dirname(os.path.realpath(__file__))
```
Stibbons: ok for now, I rework this part

Stibbons:
thanks a lot ! I'll give a try asap ! and happy new year !

PR #1051, opened 2018-01-04
update the translation system

aichingm:
* generate .desktop files for guake and guake-prefs from translations in the .po files
* implement `make update-po` to update .po files in po/
* add `make generate-mo` to generate .mo files from .po files
* add `make install-locale` to add translations to the package

additionally fix ./run-local.sh to not use the global guake package


aichingm:
Building d8fe06b on travis ci fails with `xgettext: language 'Desktop' unknown` because the gettext version used on the ubuntu 14.04 image used by travis ci is 0.18.3.1 and was released 2013-08-16. 
From what I can tell [the required version for .desktop](http://git.savannah.gnu.org/cgit/gettext.git/commit/?h=v0.19.8.1&id=f83445c64c06b41c00ebbd71b0912f1d11c63919) support is 0.19.1 which was released on 2014-06-10. 

Ubuntu 16.04 already ships with 0.19.7, but unfortunately travis ci [is still working on 16.04 support](https://github.com/travis-ci/travis-ci/issues/7260#issuecomment-354350596) 

Since I assume very few people will try to build guake on ubuntu 14.04 I will just add ` || echo "Skipping .desktop files, is your gettext version < 0.19.1?"` warnings to the build process and fallback to not translating the .desktop files. This outputs should probably be removed after travis ci supports Ubuntu 16.04.

po/ca.po
```diff
@@ -980,6 +980,128 @@ msgstr ""
...
 
+#: guake/guake_app.py:491
+msgid ""
+"Guake is now running,\n"
+"press <b>{!s}</b> to use it."
+msgstr ""
+
+#: guake/main.py:153
+msgid "Return the selected tab label."
+msgstr ""
```
Stibbons: I think we can avoid removing existing translation (for the french one, it pretty much kept all ones)

run-local.sh
```diff
@@ -13,6 +13,6 @@ EOC
 if [ \$? -eq 1 ]; then
     pew toggleglobalsitepackages
 fi
-guake --no-startup-script
+PYTHONPATH=. python3 guake/main.py --no-startup-script
```
Stibbons: why is this needed?

Stibbons:
Well this seems great! I hope we can keep the old translations. I'll see what I can do

run-local.sh
aichingm: I had guake installed from the repo, so `/usr/lib/python3.6/site-packages/guake/globals.py` existed. Then I did some changes to guake/globals.py for example 
```python
...
import sys
sys.exit(1)
...
```
and ran `make dev ; make run` guake just ran fine without exiting. Turns out that `pew toggleglobalsitepackages` also toggles the guake-site-package and the directory /usr/... has priority over ~/.local/... .
So setting the PYTHONPATH to . will set the priority back to the source directory.

po/ca.po
aichingm: I don't think that it deleted any translations. Did I miss something? The git-diff shows me only additions to the .po files but no deletions/changes.

aichingm: Ah ok. I know why we "lost" some of the translation 🤦‍♂️ the new version of the glade files does not separate combobox items by line-break but has individual xml elemets for all items (for example).

```
Never\n
With process running\n
Always
```
Versus
```
<item translatable="yes">Never</item>
<item translatable="yes">With process running</item>
<item translatable="yes">Always</item>
``` 

I have to spend some time on this! 

Info for myself: [--use-first](https://github.com/Guake/guake/blob/6a3cc349256871ff104733378f27a801182624f0/Makefile#L138) should maybe also changed.

Stibbons: Hum, after the merge the translation hasn't been lost :)

run-local.sh
Stibbons: great!

Stibbons:
merged, so we can move forward :)

Issue #280, opened 2014-03-12
Set Window title to name of active tab


hillbicks:
Hey there, 
it seems that the actual window title for guake is always set to Guake!. It would be nice to change the window title from guake to the name of the active tab. Since guake doesn't register as a visible window in the DE, it doesn't bother most people I guess, but I'm trying to interact with keepass and it's autotype function. KeePass uses the window title to identify a programm and therefore it is not possible to identify a tab, since the window title seems to be hardcoded Guake!. 

It would be nice to get such a feature. If you need further information, just let me know. 

Thanks.

naodesu:
Guake's window title can be used by time tracker called arbtt.
It can be used to provide a statistic on how much time you spent working with some program in guake or working with some host using ssh.

http://arbtt.nomeata.de/doc/users_guide/effective-use.html

Suor:
Also using time tracker (my own), meaningful window titles would be great.

Stibbons:
If you want this feature, please vote on [FeatHub](https://feathub.com/Guake/guake)

naodesu:
@Stibbons The list of features is empty, should someone add a new feature or this issue will be imported to feathub?

Stibbons:
yes, add it and put a link to this ticket if you want

PR #1063, opened 2018-01-13
update window title from active tab title

panpann:
This pull request allows to update the default window title **Guake!**, while using Guake.  

It will update it when:
- the active tab changes
- the active tab is renamed
- the vte title changes

This behavior is configurable through option *set window title to current tab name* in **Preferences** menu.

This resolves issue #280 

I'm not familiar with Guake codebase and GTK environnement. I tried to stick with the existing codebase, so let me know if there is something to change ! :blush:


Stibbons:
This looks wonderful! I integrate it in the upcoming release ! Thanks a lot, you even did the reno chunk for the release note !

panpann:
Good ! I followed your pull request guidelines :)

PR #1067, opened 2018-01-14
set reno pre_release_tag_re

dhellmann:
Configure the regex to find pre-release versions matching the style
used in this repository (ending in "\.([ab]|rc)\d*" without the "0"
before the pre-release type), rather than the default (ending in
"\.\d+([ab]|rc)\d*").

Signed-off-by: Doug Hellmann <doug@doughellmann.com>

### DIFF
--- a/releasenotes/config.yaml
+++ b/releasenotes/config.yaml
@@ -2,6 +2,7 @@
 branch: master
 earliest_version: 3.0.0.a4
 collapse_pre_releases: true
+pre_release_tag_re: (?P<pre_release>\.(?:[ab]|rc)+\d*)$
 stop_at_branch_base: true
 sections:
   # The prelude section is implicitly included.


Stibbons:
Thanks a lot for this fast patch ! I did not understood why pbr autogenerated version with alpha version with a 0 in front (3.0.0.0a2 for example)

dhellmann:
pbr is following our interpretation of https://www.python.org/dev/peps/pep-0440/#pre-releases 

I'm not sure that version numbers without the leading 0 are "wrong" they're just different than the interpretation used in pbr.

Thanks for the clear and detailed bug report, by the way. And of course let me know if I can help with reno in any other way.

Stibbons:
yes, I like reno, i think it makes a good job in building rel note for restructured. I use pandoc to convert it to github and NEWS.rst, but i have to make some kung foo. Would you be interested in several output formats for reno?

dhellmann:
Sure, I'd be happy review patches or talk about designs for ways to support alternative formats.

Issue #1113, opened 2018-01-19
[gtk3] Guake window is open upon startup


kenny-wettlaufer:
Am I missing something in the new settings window? I have a PC that runs ProjectM Visualizers & the guake terminal opens up overtop of it. I've switched back to guake-gtk2 for now. 
Sorry if this seems tedious or if I missed something but I can't have guake opening up ontop of projectM, really nice to see Guake move to gtk3!

Stibbons:
you mean guake does not open on top of ProjectM ? is it https://github.com/projectM-visualizer/projectm ? Is there other GTK3 application than can goes on top of it?

trinaldi:
I've also set guake to open on startup. When GNOME opens, Guake starts "maximized" (toggled on).

MeissnerEffect:
I have the same problem.

Stibbons:
fixed on HEAD

Issue #1111, opened 2018-01-19
2 issues - command options don't work, crash when disabling keybinding


luisnabais:
**Description**
If I run guake -t, or guake --hide, or guake --show, for example, I always get:
```
INFO:guake.guake_logging:Logging configuration complete
DEBUG:guake.guake_logging:Debug mode enabled
INFO:guake.main:Guake is already running
```
If I try to disable Keybinding for Toggle Visibility, as soon as I press Backspace, Guake crashes instantly.

i'm using Arch Linux.
This only happens in today's update, yesterday's version was working perfectly.
I tried removing the home folder for guake, but the issues remain.

I'm using GNOME, fully updated.

**Additional info**
* package version(s)
Guake Terminal 3.0.2+6+ge9df754

Steps to reproduce:
1.
Run in console/command line, any of the commands:
```
$ guake -t
$ guake --show
$ guake --hide
```
2.
Go to Guake Preferences -> Keybindings -> Try to change keybinding for Toggle Visibility and press Backspace, to remove keybinding.

Thank you very much.
Best regards,
Luís Nabais

Stibbons:
prefs panel is buggy indeed, and i already experienced you issue with "Guake is already running". DBus seems also to be not fully working

presianbg:
Guake is not starting after upgrade from 0.8 to 3.0.2 on Arch Linux.

Stibbons:
Which error?

presianbg:
```
INFO:guake.guake_logging:Logging configuration complete
DEBUG:guake.guake_logging:Debug mode enabled
DEBUG:guake.main:DBus not running, starting it
INFO:guake.guake_logging:Logging configuration complete
DEBUG:guake.guake_app:set_final_window_rect
DEBUG:guake.guake_app:  height_percents = 83.54037267080744
DEBUG:guake.guake_app:  width_percents = 100.0
DEBUG:guake.guake_app:  halignment = 0
DEBUG:guake.guake_app:  valignment = 0
DEBUG:guake.guake_app:Current monitor geometry
DEBUG:guake.guake_app:  window_rect.x: 0
DEBUG:guake.guake_app:  window_rect.y: 0
DEBUG:guake.guake_app:  window_rect.height: 1080
DEBUG:guake.guake_app:  window_rect.width: 1920
DEBUG:guake.guake_app:is unity: False
DEBUG:guake.guake_app:Correcteed monitor size:
DEBUG:guake.guake_app:  total_width: 1920
[...]

PR #1123, opened 2018-01-20
 Fix guake showing up on startup, and on flags: -t, --hide and --show 

aichingm:
Fix the first part of the issue #1111 and issue #1113 by removing testing code marked with `#TODO PORT`.


Stibbons:
thanks :)

luisnabais:
On the latest version, the issue is solved. Thanks a lot

PR #1127, opened 2018-01-21
Automatically open editor for the releasenote

aichingm:
I thought it would be nice if `make reno SLUG=<short_name_of_my_feature>` would automatically open the created file for editing for you.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -253,6 +253,7 @@ generate-desktop:
 
 reno:
 	pipenv run reno new $(SLUG)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -253,6 +253,7 @@ generate-desktop:
 
 reno:
 	pipenv run reno new $(SLUG)
+	@hash $$EDITOR 2> /dev/null; if [[ $$? -eq 0 ]]; then $$EDITOR releasenotes/notes/$$(ls releasenotes/notes -rt | tail -n 1); fi
```
Stibbons: just add `--edit` to `reno new`

PR #1133, opened 2018-01-21
Added hotkey "show-focus"

shadowxnex:
Added hotkey for showing and focusing Guake window when it is opened or closed.
It is convenient when Guake window are overlapped with another windows and user needs to just showing it without closing and opening it again.


Stibbons:
gladely accepted !

PR #1135, opened 2018-01-21
Added 'hide' hotkey for hiding guake window unconditionally without opening and showing it

shadowxnex:
The additional feature to manipulate guake window from global environment.
You can use show-hide hotkey to hide window, but for convenient purposes there are hide-win hotkey. It is useful when you don't want to check what state of guake window have and just to use direct action.

As an addition to #1133


Stibbons:
can you rebase please?

aichingm:
Maybe a prescription in the  guake/data/org.guake.gschema.xml file would be helpful.

Stibbons:
and a small `make style check` as well for travis :)

gsemet:
can you rebase and update the PR or I'll close it. Thanks

PR #1137, opened 2018-01-21
 Rewrite hotkey management in pref panel 

Stibbons:
+ default history-size set to 1000


Stibbons:
i do not promise this is bug-free but at least it works :)

aichingm:
c2abe11: I left the visible bell checkbox there so that `make pot` will keep the translations in the .po files. Which was in my eyes saving some work for reimplementing the vissible bell feature.

Stibbons:
indeed. i revert this and simply remove the event method

PR #1141, opened 2018-01-21
Add the possibility to terminate guake with ctrl+c

aichingm:
This allows to terminate guake with ctrl+c when running via `make run` or from command line.

### DIFF
--- a/releasenotes/notes/add-control-c-handler-fb2967d1ba8ee495.yaml
+++ b/releasenotes/notes/add-control-c-handler-fb2967d1ba8ee495.yaml
@@ -1,0 +1,3 @@
+features:
+  - |
+    Add the possibility to terminate guake with ctrl+c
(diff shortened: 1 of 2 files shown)


releasenotes/notes/add-control-c-handler-fb2967d1ba8ee495.yaml
```diff
@@ -0,0 +1,3 @@
+features:
+  - |
+    Add the possibility to terminate guake with ctrl+c
```
Stibbons: can you add "from the starting console, so it won't affect ctrl+c from inside guake's terminals" or something like that?

aichingm: I'm sorry I have been on a little trip the past few days. Thanks for making the change  👍 

> Add the possibility to terminate guake with `Ctrl+c` on terminal where Guake has been launched

Sounds good!

Stibbons: no pbl

PR #1164, opened 2018-01-25
 Allow make install-system to be run as non root user

### LINKED ISSUE #1139: Running make instal-system as non root user
I using `make install INSTALL=*** PREFIX=***` to build packages for archlinux as none root user which works perfect with one little problem: 

https://github.com/Guake/guake/blob/41c5b8b408b0360483f2e467f616f88a534acf83/Makefile#L44

needs root and fails if I try to run it. Would it be ok to replace this line with something like:

    @update-desktop-database || echo "Could not run update-desktop-database, are you root?"
Or should make install-system only be used to install guake in /usr/...?

This would also be very useful if one would like to install guake on a system where he/she has no root access. (eg for installing guake in ~/.local/...)

aichingm:
This closes/fixes #1139


Stibbons:
thx !

PR #1167, opened 2018-01-25
Add window displacement options to move guake away from the screen edges

### LINKED ISSUE #1162: Dual stacked monitor configuration misposition
I'm using two 1080p monitors placed one over another. And when Guake located on bottom one - which is with a top bar - it's overlapped with a top bar. Details in the attached screenshot.
screenshot from 2018-01-24 15-57-07

aichingm:
This PR adds window displacement options to move guake away from the screen edges and fixes #1162. BTW it lets you go all crazy about the positioning of Guake:

screenshot_20180125_141027


Confucij:
vim swap file in commit - is it mistake?

Stibbons:
gladly accepted!

aichingm:
@Confucij yes I deleted it, I accidentally added it in an earlier commit. Maybe *.swp would be a good idea for .gitignore.

Confucij:
Perfect! 3.0.4 fixed all my issues.

Confucij:
@aichingm  My joy wasn't too long( Positioning is same on both monitors so it's good on lover display and 25px lover on top monitor(

Stibbons:
hum if you can understand where this 25px comes from it can help

Confucij:
`vdisplacement` applies on each monitor but 25px-height-panel only available on bottom one. I think the only way to fix this is to add configuration for each monitor in the system. And I don't think this should be prior task - stacked monitors not so widespread in the wild.

PR #1169, opened 2018-01-25
fix preferences window header color, align close button and change borders to margins

aichingm:
From:

screenshot_20180125_212018

To:

screenshot_20180125_211946


aichingm:
Do we even need a close button?

Stibbons:
i prefere keeping the close button, i don't rely like to rely on the x button. at least this is a part of most ui on macoc that always annoys me. maybe i am old school or used too much windows...

Stibbons:
looks good

releasenotes/notes/fix-preference-window-header-color,-align-close-button-and-change-borders-to-margins-fa7ffffc45b12ea5.yaml
```diff
@@ -0,0 +1,29 @@
+fixes:
+  - |
+    fix preferences window header color, align the close button more nicely and change borders to margins
+
+translations:
+    - pt_BR
+    - sv
```
Stibbons: i have reverted this part, these translations have not been updated (or if you speak 20 languages?), just mo files are generated. translations => something updated the translation

aichingm: No I actually changed all of them by hand (poedit) I removed some markup which was possible without knowing the language. You can remove it from the releasenote but the po files are changed by me (all of them)

Stibbons: yes of course i keep the change in po. thanks for that !

Issue #1170, opened 2018-01-25
Preferences window launcher on none Gnome systems


aichingm:
This line disables the Preferences launcher for all none Gnone systems:
https://github.com/Guake/guake/blob/93099961f7c90a22089b76a8a9acf1414bea56e5/guake/data/guake-prefs.template.desktop#L10

Is there a reason for this?

Can it be removed? Or should we add an separate one for none Gnone systems? With something like: `NotShowIn=GNOME` in it?

Stibbons:
can you try removing it? I actually added it in an attempt to restore it, but in either case it does not work :(
ubuntu 17.10 and up disallow custom setting in system panel :(

PR #1180, opened 2018-01-27
fix preferences desktop file

aichingm:
`TryExec` should only be the file name see https://developer.gnome.org/desktop-entry-spec/
`OnlyShowIn` removed

Closes: #1170


Stibbons:
thanks. i'll give a try

Stibbons:
works great !

Issue #1189, opened 2018-02-11
'make run' fail 


bronvic:
OS: Arch linux x86_64 4.15.2-2-ARCH
GNOME Shell: 3.26.2

HTR:
1. Clone **latest** master
1. run *make run*
1. :boom: observe this:
```
python/guake [master]$ make run
rm -f guake/data/gschemas.compiled
glib-compile-schemas --strict guake/data/
export GUAKE_DATA_DIR=/home/voronwe/projects/python/guake/data ; pipenv run ./run-local.sh
execute Guake GTK3 for developer.
Logging configuration complete
DBus not running, starting it
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/dbus/bus.py", line 175, in activate_name_owner
    return self.get_name_owner(bus_name)
  File "/usr/lib/python3.6/site-packages/dbus/bus.py", line 361, in get_name_owner
    's', (bus_name,), **keywords)
  File "/usr/lib/python3.6/site-packages/dbus/connection.py", line 651, in call_blocking
    message, timeout)
dbus.exceptions.DBusException: org.freedesktop.DBus.Error.NameHasNoOwner: Could not get owner of name 'org.guake3.RemoteControl': no such name

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "guake/main.py", line 241, in main
    remote_object = bus.get_object(DBUS_NAME, DBUS_PATH)
  File "/usr/lib/python3.6/site-packages/dbus/bus.py", line 241, in get_object
    follow_name_owner_changes=follow_name_owner_changes)
  File "/usr/lib/python3.6/site-packages/dbus/proxies.py", line 248, in __init__
    self._named_service = conn.activate_name_owner(bus_name)
  File "/usr/lib/python3.6/site-packages/dbus/bus.py", line 180, in activate_name_owner
    self.start_service_by_name(bus_name)
  File "/usr/lib/python3.6/site-packages/dbus/bus.py", line 278, in start_service_by_name
    'su', (bus_name, flags)))
  File "/usr/lib/python3.6/site-packages/dbus/connection.py", line 651, in call_blocking
    message, timeout)
dbus.exceptions.DBusException: org.freedesktop.DBus.Error.ServiceUnknown: The name org.guake3.RemoteControl was not provided by any .service files

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "guake/main.py", line 368, in <module>
    exec_main()
[...]

PR #1190, opened 2018-02-11
Fixed path to guake data in makefile

bronvic:
#1189 

The problem is: make runs from *guake* folder and data stores into *guake/guake*.
I see two solutions:
1. Edit *GUAKE_DATA_DIR* into Makefile, as I do or
1. Remove export of *GUAKE_DATA_DIR* from Makefile, because globals would set correct [default value](https://github.com/Guake/guake/blob/master/guake/globals.py#L52) if *GUAKE_DATA_DIR* is not in environmental variables

First one seems better for me because someone can set this env var manually

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -138,7 +138,7 @@ wheels:
 
 
 run-local: compile-glib-schemas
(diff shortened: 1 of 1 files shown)


gsemet:
looks good to me! thanks!

PR #1214, opened 2018-03-22
Fix cursor blinking mode in new tabs

phihag:
Previously, cursor blink mode was just applied when changing.

Steps to reproduce the issue:

1. Open guake
2. Change Cursor Blink Mode to Off in Preferences -> Appearance
3. Observe: Blink mode of current terminals is correctly changed, no more cursor epilepsy.
4. Open a new terminal (Right-click -> New Tab)
5. Observe: Blinking cursor, setting is not applied.

The fix is trivial: Upon tab start, apply the blink mode property.


gsemet:
great thankx!

PR #1251, opened 2018-04-10
Rework install system following packagers recommendations

gsemet:


gsemet:
@aichingm I have reworked a little bit the packaging and localization (works on ubuntu now perfectly!), can you test it on arch?

aichingm:
testing...

aichingm:
Looks good so far. The only thing which would make the life of packagers a bit more easy is if there were an option not to create `/usr/share/glib-2.0/schemas/gschemas.compiled` (and there for exclude it from the build) in arch this file interferes with the existing compiled schemas file and recompiling it is handled by the system.  But this is only a minor thing. 

With my last pr my build file is already down to 2 LOC (`make install` and `rm /.../usr/share/glib-2.0/schemas/gschemas.compiled`) which is awesome :+1: :tada:

gsemet:
why not optionalize this option in the makefile (COMPILE_GSCHEMA)?

PR #1254, opened 2018-04-11
fixes for the build system (split INSTALL_ROOT from PREFIX)

aichingm:
`INSTALL_ROOT` should not be part of `PREFIX`. Eg.: if you want to build guake so you can run it from /usr/bin (not /usr/local/bin) you set the `PREFIX` to "/usr" but you also want to build it in a different directory then /usr (eg for packaging) than you set the `INSTALL_ROOT` to /home/.../guake/build/.This will fail since `PREFIX` will ignore the `INSTALL_ROOT` and your python files will end up in the right dir because of `@$(PYTHON_INTERPRETER) setup.py install --root "$(INSTALL_ROOT)" --prefix="$(PREFIX)" --optimize=1` but `make` will try to install the data files in /usr where they should not be. 

By separating the `INSTALL_ROOT` from the `PREFIX` you can "jail" the build to a particular root directory and still choose a different `PREFIX`.

I hope this somehow seams understandable :face_with_head_bandage: if not please ask! 

PS: this would remove a lot of `sed` magic from the build proccess. 

In response to #1251

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 PYTHON_INTERPRETER=python3
 MODULE:=guake
 INSTALL_ROOT:=/
-PREFIX:=$(INSTALL_ROOT)usr/local
+PREFIX:=/usr/local
 # findout the dist-package directory name. Under most system, users can install on dist-package.
 # On debian, site-package is reserved for the official python packages and has precedence over
 # dist-package
@@ -77,7 +77,7 @@ install-locale:
 	for f in $$(find po -iname "*.mo"); do \
 		l="$${f%%.*}"; \
 		lb=$$(basename $$l); \
-		install -Dm644 "$$f" "$(LOCALE_DIR)/$$lb/LC_MESSAGES/guake.mo"; \
+		install -Dm644 "$$f" "$(INSTALL_ROOT)$(LOCALE_DIR)/$$lb/LC_MESSAGES/guake.mo"; \
 	done;
 
 uninstall-locale: install-old-locale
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -87,33 +87,35 @@ install-old-locale:
...
+	install -Dm644 guake/data/pixmaps/guake.png "$(INSTALL_ROOT)$(PREFIX)/share/pixmaps/guake.png"
+	mkdir -p $(INSTALL_ROOT)$(SHARE_DIR)
+	mkdir -p $(INSTALL_ROOT)$(GLADE_DIR)
+	install -Dm644  guake/data/*.glade "$(INSTALL_ROOT)$(GLADE_DIR)"
+	mkdir -p $(INSTALL_ROOT)$(SCHEMA_DIR)
+	install -Dm644 "guake/data/org.guake.gschema.xml" "$(INSTALL_ROOT)$(SCHEMA_DIR)/"
 
 compile-shemas:
-	glib-compile-schemas $(PREFIX)/share/glib-2.0/schemas/
+	glib-compile-schemas $(INSTALL_ROOT)$(PREFIX)/share/glib-2.0/schemas/
```
jtojnar: This should use `$(INSTALL_ROOT)$(SCHEMA_DIR)`

gsemet: please use the prefix for that. It should have the root inside it already

gsemet:
ok I see the reason

Makefile
aichingm: Yes good catch. But is it needed to build guake? shouldn't  make compile-schemas only be used if you actually install guake in $(PREFIX)? how is this handled in nix, in arch pacamn runs glib-compile-schemas every time you install a new file in /usr/share/glib2.0./...

gsemet: I saw one packager (arch?) that delete this line, don't know why. I need to compile it to debug (dev env) or from /usr/local/... from my ubuntu

jtojnar: The concept is not very compatible with Nix’s use of separate immutable prefix for each package. Moreover, Nix also has *profiles*, which are just package directory trees merged into a single tree via `lndir`, simulating FHS. This merger would create a [conficts](https://github.com/NixOS/nixpkgs/issues/1455) between `share/glib-2.0/schemas/gschemas.compiled` files from different packages, so for now, we compile it for each package separately during installation (often handled by makefiles or `meson_post_install` script), and then move the file to `share/gsettings-schemas/${package-name}-${package-version}/glib-2.0/schemas/`.

aichingm: @gsemet pacman (the package manager in arch) runs a hook after the install process which handles the schema compilation. The line is probably removed because during packaging recompiling is not needed.

jtojnar: Yes, the compiled file depends not only on the files from the project, but also on everything else that is in the directory. For that reason, it should not be a responsibility of project build scripts (Makefiles) but one of package manager (as pacman does with post-install hooks).

Unfortunately, this falls short when one does not use a package manager (e.g. developer installing the app to `/usr/local/share` for testing), so most project do have `glib-compile-schemas` somewhere in their build scripts. 

One possible “solution” is building the caches (schemas, indexes, etc.) only if `DESTDIR` environment variable [is not set](https://github.com/GNOME/gnome-shell/blob/9f7b1014378a756f761908cc44da0ce07a35ab9d/meson/meson-postinstall.sh) and relying on the fact that most package managers set it.

aichingm: Nice idea! I thought of something more explicit like an opt-out flag (less "magic").

gsemet: I leave that to you, do not hesitate to add this opt-out flag

aichingm: @gsemet isn't that why you added COMPILE_SCHEMA? This works for me on arch.

gsemet: If it is enough for you it’s perfect !

PR #1256, opened 2018-04-11
 Add "start at login" option + standart make names

gsemet:


gsemet:
@aichingm Can you try after this rework? I hope this would make everybody happy!

aichingm:
I'm on it.

aichingm:
Works like a charm with #1258.

PR #1296, opened 2018-04-21
Fixed "Gruvbox Dark" palette

lonepie:
Foreground and Background needed to be reversed.


gsemet:
Thanks! Can you add a note in the release note ("make reno SLUG=palette") so that you get credited for this fix?

PR #1301, opened 2018-04-25
Fix url match

### LINKED ISSUE #1295: Open URL ceased to work
Hi,

I'm running Arch, I've just updated guake to 3.2.0, with gnome 3.28.1

There's also been a python update with the same system update, so *maybe* I've just switch from gtk2 version to gtk3, I don't know for sure. Maybe I have remnant of 0.8 guake somewhere in a conf file, I did notice that I had to redo some manual config, like global dark theme. So ... maybe there's a remnant of config somwhere in the way.

## Before :
Until then, I could 'ctrl click' on anything that looked like an url  (say : https://github.com/Guake), it would open it in my default browser. The link itself was disaplyed with an underline.

## After :
Now, it doesn't work anymore
No underline, nothing happens on ctrl click.

When launching guake from a terminal I can see warnings, but nothing once it's launched. FOr the records here are the warnings :

`
(guake:4380): Vte-WARNING **: 15:28:05.004: (vtegtk.cc:1916):int vte_terminal_match_add_regex(VteTerminal*, VteRegex*, guint32): runtime check failed: (_vte_regex_get_compile_flags(regex) & PCRE2_MULTILINE)
`

It's 100% reproductible on my PC, even after reinstall. Did not purge entirely config files ... since I don"t know which one I should delete. I can try if someone gives me the list.

regards,
and thanks for the job !

jaawerth:
Updated `terminal.py` to use the [documented function signature](https://lazka.github.io/pgi-docs/Vte-2.91/classes/Regex.html#Vte.Regex.new_for_match) for `Vte.Regex.new_for_match`. Tested locally, and this corrected the broken hyperlink matching in `3.2.0`. Let me know if I missed anything!

Fixes #1295

### DIFF
--- a/guake/terminal.py
+++ b/guake/terminal.py
@@ -170,11 +170,11 @@ class GuakeTerminal(Vte.Terminal):
         """
         try:
             for expr in TERMINAL_MATCH_EXPRS:
-                tag = self.match_add_regex(Vte.Regex.new_for_match(expr, 0, 0), 0)
+                tag = self.match_add_regex(Vte.Regex.new_for_match(expr, len(expr), 0), 0)
                 self.match_set_cursor_type(tag, Gdk.CursorType.HAND2)
 
             for _useless, match, _otheruseless in QUICK_OPEN_MATCHERS:
-                tag = self.match_add_regex(Vte.Regex.new_for_match(match, 0, 0), 0)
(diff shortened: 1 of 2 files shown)


gsemet:
thansk!

PR #1302, opened 2018-04-25
minor fixes

gsemet:


docs/source/contributing/dev_env.rst
```diff
@@ -2,20 +2,49 @@
...
+    $ ./scripts/bootstrap-dev-fedora.sh
+
+Arch Linux
+----------
+
+Execute:
+
+.. code-block:: bash
+
+    $ ./scripts/bootstrap-dev-arch.sh
```
aichingm: Did I miss something or does this file `./scripts/bootstrap-dev-fedora.sh` not exist?

gsemet: yes it is on my other branch at home I'll upload it tonight :)

aichingm: Didn't saw you comment. Forget my second ;)

gsemet: perfect I'll use it!

aichingm: :laughing: Ok

gsemet:
@aichingm thanks for your help. Do you think guake works quite properly on Arch for the moment? Do you think I should release 3.2.1?

aichingm:
Yes I think so. I mean there are the issues you know about like quickopen and url links but I don't think that there are more arch specific issues, building works like a charm, and for me it is running without any show-stoppers. Except one which I just found (minor, I will open an issue) by clicking through the preferences window  and testing. Even the 'Start Guake at login' works which I find is awesome!

gsemet:
cool. That's odd because I have quick open and hyperlinks working on fedora and ubuntu. I'll release we'll see that for next version :) thanks for the help!

Issue #1198, opened 2018-02-18
Ctrl+D in the terminal does not close the tab anymore, but freezes the current tab


XaF:
## Version information
**Guake version:** 3.0.4
**Linux version:** Debian buster
**Gnome version:** GNOME Shell 3.26.2

## Expected behavior
When in a terminal, using Ctrl+D on an empty prompt exits the current terminal.
In previous versions of Guake, this would act as if the tab was closed: if it's not the last tab, only this tab is closed, if it's the last tab, the full Guake window would hide until the next F12 key press.

## Current bug
When using Ctrl+D on an empty prompt, the current tab freezes after printing "exit" on the terminal. The context menu can be opened on a right click, but selecting any option does not do anything. If another tab is opened, it can be used to right click but not all options work: 'New tab' works, 'Quit' does not work (probably blocked by the other tabs). Only solution to get back full control of Guake is to `killall guake`.

romange:
The same also happens in Ubuntu 18.04

gsemet:
I do not have this behavior. Ctrl+d on a gnome tab close nicely this tab. Can you start guake from a terminal and dump me the logs if you see error message? Can you try with HEAD? (cannot reproduce on Guake 3.0.5)

romange:
I am using guake 3.0.4-1 coming from ubuntu repo.
Below are the logs printed once I typed "exit" in my guake shell. But the tab hangs like in the attached screenshot.

![screenshot from 2018-03-20 12-50-15](https://user-images.githubusercontent.com/3674760/37650326-4dc2b42e-2c3d-11e8-9f73-ea49f8e85b01.png)


```
Received window state event: <flags GDK_WINDOW_STATE_STICKY | GDK_WINDOW_STATE_ABOVE | GDK_WINDOW_STATE_FOCUSED | GDK_WINDOW_STATE_TOP_RESIZABLE | GDK_WINDOW_STATE_RIGHT_RESIZABLE | GDK_WINDOW_STATE_BOTTOM_RESIZABLE | GDK_WINDOW_STATE_LEFT_RESIZABLE of type Gdk.WindowState>
Terminal exited: <terminal.GuakeTerminal object at 0x7f0ab03e4ee8 (guake+terminal+GuakeTerminal at 0x186b440)>
[...]

PR #1313, opened 2018-04-27
add option groups (run, make, dev, opt) to the bootstrap scripts

aichingm:
I added options to the scripts so that users who just want to install the dependencies needed for running and making guke (`make install`) like here <https://github.com/Guake/guake/issues/1198#issuecomment-385054211> can install dependencies more easily.

I basicly grouped the dependecies in to 4 groups 
* run: packages needed for running
* make: packages needed for `make && make install`
* dev: packages like poedit which are just needed for developing
* opt: optional packages for running 

```
./scripts/bootstrap-debian.sh # will install every group except opt

./scripts/bootstrap-debian.sh run # will install only the run group
./scripts/bootstrap-debian.sh dev # will install only the dev group

./scripts/bootstrap-debian.sh run make # will install the run and the make group
./scripts/bootstrap-debian.sh make run # will install the run and the make group

./scripts/bootstrap-debian.sh make run opt # will install the run, make and opt group

...
```
I also added a paragraph about the dependencies to <http://guake.readthedocs.io/en/stable/user/installing.html#install-from-source> (docs/user/install.rst)

### DIFF
--- a/docs/source/user/installing.rst
+++ b/docs/source/user/installing.rst
@@ -41,7 +41,16 @@ your system.
 Install from source
 ===================
 
-If you want to install Guake from its sources, use:
+If you want to install Guake from its sources make sure that you have the 
+needed dependencies installed. If you are unsure about the dependencies you 
+can run to install them:
+
+.. code-block:: bash
+    $bash script/bootstrap-[debian, arch, fedora].sh run make
+
+Note: inster your distribution in the square brackets.
+
+To install Guake itself, use:
 
 .. code-block:: bash
 
--- a/scripts/bootstrap-dev-debian.sh
+++ b/scripts/bootstrap-dev-debian.sh
@@ -1,36 +1,76 @@
 #!/bin/bash
 
-echo "Install packages needed for execution"
+RUN=0
+MAKE=0
+DEV=0
+OPT=0
+ARGC="$#"
 
-sudo apt install -y \
-    gir1.2-keybinder-3.0 \
-    gir1.2-notify-0.7 \
-    gir1.2-vte-2.91 \
-    libkeybinder-3.0-0 \
-    python3 \
-    python3-cairo \
-    python3-dbus \
-    python3-gi \
-    python3-pbr \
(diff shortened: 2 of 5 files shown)


scripts/bootstrap-dev-debian.sh
```diff
@@ -1,36 +1,76 @@
...
+          RUN=1;
+          MAKE=1;
+          DEV=1;
+          OPT=0;
+fi
+
+
+if [[ $RUN == "1" ]]; then
+    echo "Install packages needed for execution"
     sudo apt install -y \
```
aichingm: @gsemet do you need the -y flag or is it just for travis? If it is just for travis could we remove the flag from this file and add `yes | sudo ./scripts/bootstrap-dev-debian.sh` to .travis.yml.

aichingm: oh and by the way same question for sudo.

gsemet: I am ok if you update the Travis script as well. I have no problem with sudo in a script.

docs/source/user/installing.rst
```diff
@@ -41,7 +41,16 @@ your system.
...
 
-If you want to install Guake from its sources, use:
+If you want to install Guake from its sources make sure that you have the 
+needed dependencies installed. If you are unsure about the dependencies you 
+can run to install them:
+
+.. code-block:: bash
+    $bash scripts/bootstrap-dev-[debian, arch, fedora].sh run make
+
+Note: inster your distribution in the square brackets.
```
gsemet: Insert

aichingm:
Do you mean just having one script which covers all distributions (I could do that)? It sure it's nice for packagers to have such a file but the last issues to which I have responded to had trouble running `make install` because of missing dependencies...

scripts/bootstrap-dev-debian.sh
aichingm: Ok I will push an update to this pr tomorrow.

PR #1355, opened 2018-07-07
Don't translate application icon

rzimmer:
As there is no locale-specific application icon, there is no need for having translation values for them.

This commit removes all unnecessary application icon translations resulting in clearer *.desktop files.
Besides, this finally fixes the Guake application icon not displayed with German locale as "de.po" put
a wrong icon name into the generated *.desktop files.

The already merged PR #1320 only fixed the .desktop-file for autostart, but not for the application launcher from the desktop environment - sorry about that!


gsemet:
Hi. Can you rebase this branch so I can merge it ? Thanks for it

rzimmer:
Of course, done now :)

gsemet:
and merged!

PR #1356, opened 2018-07-08
Completely overhaul German translation

rzimmer:
`de.po` is re-created based on a freshly generated `guake.pot` file, so the translation file is cleaned from obsolete or duplicated translation strings. All missing German translation strings (mainly for newer options) have been added. As a bonus, the label `Set window title to current tab name` is now translatable (and already translated to German), too!

The overhauled translation follows the German translation guidelines by GNOME (https://wiki.gnome.org/de/UebersetzungsRichtlinien) as best as possible. This improves wording consistency with other GNOME applications, e.g. GNOME Terminal.

Special care was taken to be as close as possible with the meaning of the English translation to avoid ambiguities. Particularly, preferences are explained clearer to the user now.

Besides, a whole lot of typos and punctuation fixes are included.

The updated translations in screenshots:

context_menu_de

preferences_01_de

preferences_02_de

preferences_03_de

preferences_04_de

preferences_05_de

preferences_06_de

preferences_07_de

preferences_08_de

preferences_09_de

about_01_de

about_02_de

about_03_de

PS: Didn't dare to update the translation credits yet, but I'll be more than happy to do so if my translation effort gets merged :smile:

### DIFF
--- a/data/autostart-guake.desktop
+++ b/data/autostart-guake.desktop
@@ -1,7 +1,7 @@
 [Desktop Entry]
 Name[ca]=Terminal Guake
 Name[cs]=Guake Terminal
-Name[de]=Guake-Terminal
+Name[de]=Guake Terminal
 Name[es]=Guake Terminal
 Name[fa]=Guake Terminal
 Name[fr]=Terminal Guake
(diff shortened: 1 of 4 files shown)


gsemet:
Hi ! Of course your effort will be rewarded... by all my gratitude :) many thanks for all your hard work, do not hesitate to update the credit in the reno release note and in the credits !

PR #1357, opened 2018-07-10
Update about screen

rzimmer:
Remove `translatable` attribute from `copyright`, `website_label`,
`translator_credits` and `license`.

While it does not make too much sense to translate them, it definitely makes
maintaining above data much harder as there is not a single place to add or
modify it but distributed over many language files, gets easily forgotten
to update and thus leads to inconsistent information.

Furthermore, translator credits were updated by adding *all* missing translation
authors and sorting them first by (English) language name and second by the
author's forename.


rzimmer:
Additional note: Thought about removing the `translatable` attribute from `license`, too, but I'm not quite sure about that one. What do you think?

gsemet:
agree. Please remove the translatable att of licence.

rzimmer:
Great! I just updated the PR accordingly :smiley:

piotrdrag:
Why don’t you just use [translator-credits](https://developer.gnome.org/gtk3/stable/GtkAboutDialog.html#gtk-about-dialog-set-translator-credits) and [license_type](https://developer.gnome.org/gtk3/stable/GtkAboutDialog.html#gtk-about-dialog-set-license-type)?

rzimmer:
@piotrdrag As the about screen currently is a custom implementation and I didn't want to dive deeper, I just updated the language files and `about.glade`. But you're right that this would have been the appropriate and cleanest way to implement the about screen using Gtk.

I definitely see switching to `GtkAboutDialog` as a possible future PR, but at least I won't find the time to try to implement this change anytime soon unfortunately...

gsemet:
This is not a priority, but of course pull request are accepted.

PR #1369, opened 2018-07-31
remove rope dependency

### LINKED ISSUE #1366: Drop dependency on rope; CVE-2014-3539 fix pending 4 years now
Hey, @gsemet

I see you're doing great job maintaining Guake. Can you please look if it's possible to get rid of the dependency on the `rope` module? Because of this: python-rope/rope#105 -- years have passed, no RCE fix merged still.

Even though it's only a dev-dependency -- I guess nobody wants to get hacked while working on a small feature/fix to Guake just because we couldn't get out crap together.

cc @pypingou @clarete

gsemet:
Fix #1366

+ freeze pip and pipenv


gsemet:
merged

Issue #1375, opened 2018-08-15
Why does guake has ists own tabs implementation


aichingm:
Hey @gsemet,

is there a reason why guake does not use the Gtk.Notebook's tab implementation but a custom radio button group?

cheers Mario

gsemet:
Hi. Not sure why, probably a very old code. I am about to rewrite this part to enable much expected feature, but this part is kind of ugly I need to break everything :(

aichingm:
Are you working on it right now? I just started to work on split terminal functionality, and rewriting the notebook part.

aichingm:
The things I thought about doing are: 
1. remove the custom tabs and use the notebook's
1. resolve `# TODO PORT ...` sections (finally)
1. rewrite the notebook to allow multiple terminals on one page and remove the term_list 
1. also move the functions rename_tab, delete_tab and add_tab to the notebook
1. implement split terminal using Gtk.Paned (I know this is not as flexible as the split functionality of terminator but I see no point in reimplementing terminator, that would be a hell of a task) and a bunch of recursive calls on the widget hierarchy . 

Do you have any ideas, opinions or inputs?

gsemet:
I mostly agree with you:
1. => yes, if you are really motivated to do so,  feel free.
2. => could be done in another task, couldn't it? the mosly awaited one being tab reorg
3. => sure !
4. => yes, that would simplify a lot futur evolutions
5. => I agree with you that terminator (and Tilix) does a pretty good job. But I would like to have something elegant and efficient. Being able at least to split another split (ex: vertical split the screen + the terminal on top being splitted horizontally). Ok to use Paned first but with the option later to switch to a custom made widget (derived from paned?) if the need is. 

For me, the real added value for guake would be to have easy navigation between these tiles. I always found it hard to use on other software (even on iterm2) so if Guake can do less extravagant features but do its set of feature right, I think we will be really happy with it!

gsemet:
yes, let us forget about drag n drop splited terminal for now. My guess is that should be feasible with Paned.

aichingm:
Hey @gsemet,

I have some logistical things to ask. Since this are a lot of changes (I might accidentally break things...) and not only a few bug fixes I would like to open a branch (via PR) on which I can push code via PR so that it can be tested before it is merged into master. Do you have a naming schema you want me to follow? Something like "big-rewrite", "devel", "version-x.x.x" etc...

[...]

Issue #195, opened 2013-06-27
Scrollable tab bar 


kikecalpe:
Scrollable tab bar http://guake.org/ticket/65
Reported by:    rooty   Owned by:   somebody
Priority:   major   Milestone:  0.5.0
Component:  guake   Version:    0.4.4
Keywords:       Cc:     gualo82@…, thomas.bouffon@…
Description

Tab bar should become scrollable when tabs exceed the screen width. To reproduce, try creating lot of tabs.

ivoshm:
+1

r-barnes:
Guake 0.8.4 seems to have a scrollbar in it? Could we close this issue?

Unfortunately, that scrollbar also overlaps the tabs, so I've just been looking for a way to remove it from my local copy.

Hijis:
Hi! Seems this is right place for my request.
Usualy I have too much tabs so scrollbar appears, and switch between them using Ctrl+PgUp|PgDown
But when I switch to tab beyond visible space - happens nothing and I loose tab out of my sight.

It would be nice if scrolls automaticly to current active tab. 

Thanks a lot!

![guake_tabs](https://user-images.githubusercontent.com/17022100/29557442-c71e9498-8731-11e7-9d24-1574431e64dd.png)

aichingm:
This issue is fixed via #1382

PR #1382, opened 2018-08-21
Switch to the notebook's tabs and reenable reordering

aichingm:
Sorry for the delay, had some work related thing to finish up.

I have not jet added a slug but I wanted to publish the code so that someone could have a look at it.

Also fixes: #1375


aichingm:
this should also close #195

PR #1384, opened 2018-08-23
Fix a lot of "TODO PORT" sections

aichingm:
There are still some `TODO PORT` sections left but I can't test stuff related to Unity and appindecator since I don't have a 16.-- Ubuntu box running...

I ported the monitor selection when using the "use_mouse" option to Gdk since the Gtk parts  are deprecated, but some one should test them (I have currently only my laptop with me and no access to a second monitor) 

I also changed https://github.com/Guake/guake/compare/master...aichingm:dev/todo-port?expand=1#diff-7a93ec37877f9e75dacff02a373a6734R39

and added a `run-prefs` make target to make development of the prefs window more enjoyable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -213,6 +213,9 @@ wheel: wheels
 run-local: compile-glib-schemas-dev
 	pipenv run ./scripts/run-local.sh
 
+run-local-prefs: compile-glib-schemas-dev
+	pipenv run ./scripts/run-local-prefs.sh
+
 run-fr: compile-glib-schemas-dev
 	LC_ALL=fr_FR.UTF8 pipenv run ./scripts/run-local.sh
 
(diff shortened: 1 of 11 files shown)


guake/paths.py
```diff
@@ -0,0 +1,98 @@
+# -*- coding: utf-8; -*-
```
gsemet: normally you do not commit this file is automatically generated by the makefile

gsemet:
look wonderful ! I'll test dual monitoring later this week or next week, but if it is broken i'll fix it, do not worry.

Can you add a `make reno SLUG=gtk3` about your work on all these ports? You can fill it in several commits when you submit your various pull requests !

guake/paths.py
aichingm: Of course, you are right!
I rebased and removed the file.

aichingm:
I have already done the rewrite of the notebook to allow for multiple terminals in one page, just waiting for this to be merged to keep the PRs small. It is not the most beautiful code jet since this is only step 3. and not all relevant functions are moved from guake_app.py to notebook.py but its a good start.

gsemet:
thanks a lot !

PR #1385, opened 2018-08-24
Rewrite the notebook to support multible terminals in one page

aichingm:
This is stage 3 from #1375 

I have not jet moved all code from guake_app.py to notebook.py but this is planed for stage 4. the main things to say are there is no more `term_list` if you need to get the terminals from the notebook check out the recursive calls `get_terminals` and `and iter_terminals`. Another change is `that get_current_terminal` does not return the first terminal of the current active page anymore but the last focused terminal of all terminals in the notebook.

The structure of the notebook widget hierarchy is:

```
                                  TerminalNotebook
                                /        |         \
                          page          page       page  # a page is a GtkWidget in this case a RootTerminalBox which is a Gtk.Box
                        /                 |             \
             DualTerminalBox        TerminalBox         TerminalBox
                /       \                 |                  |
    TerminalBox        TerminalBox    GuakeTerminal     GuakeTerminal
         |                  |
  GuakeTerminal      GuakeTerminal
``` 
All three `RootTerminalBox`, `TerminalBox` and `DoalTerminalBox` are of the type `TerminalHolder` checkout its methods to get an idea how and what recursive calls are supported.
  
`RootTerminalBox` and `TerminalBox` are `Gtk.Box`s.
`DualTerminalBox` is a `Gtk.Paned`

A `TerminalBox` and "its" `GuakeTerminal` should never be separated once the terminal is set on to the TerminalBox (TerminalBox.set_terminal()). 

If a terminal needs to be split call split(...) on the `TerminalBox` this will remove it self from its parent `TerminalHolder` and add a `DualTerminalBox` to the parent to which it (the TermianlBox) will add itself. *This is not jet implemented.* *Note to the person how implements this `last_terminal_focused` will need some rework too.*

No references are kept to the terminals it self it is all handled via the GtkWidget hierarchy, with one exception the `last_terminal_focused` field in `RootTerminalBox` and `TerminalNotebook`. 

Note the new code is not documented at the moment but I guess it is pretty self explanatory.

In the current state guake works as expected (like before).


gsemet:
look great !

gsemet:
I do not have time to test it this week end, but this look very serious !

Issue #1363, opened 2018-07-22
Copy URL to clipboard in right-click menu


spvkgn:
In Guake 0.8 have been context menu entry for copy URL to clipboard, in Guake 3 I did not find this feature which would be useful. At least upper 'Copy' entry does not copy URL to clipboard.

Guake 3.2.2
Ubuntu 18.04, MATE Desktop 1.20, gtk 3.22.30, vte 2.91 (0.52.2)

gsemet:
You need to select and copy the url

aichingm:
@spvkgn What would be the difference between `copy` and `copy URL`?
I can't remember to have ever used `copy URL` so pardon my asking.

spvkgn:
@aichingm I mean that in 0.8 the `Copy` allows to copy URL to clipboard easier - without selecting full text with mouse as it needs in 3.x.

gsemet:
Indeed, this has been deprecated. Copy without selection means copy the current word, copy with selection will of course copy the selected text.
There is the "Open link" conditionally displayed if the cursor is on a link, maybe we can add a "copy hyperlink" item in the menu

PR #1390, opened 2018-08-31
context menus per terminal / move code around

### LINKED ISSUE #1370: Settings schema 'guake.general' does not contain a key named 'display_n'
Whenever I try to open guake preferences (`guake -p`), I get:

`(guake:2531): GLib-GIO-ERROR **: 16:26:13.476: Settings schema 'guake.general' does not contain a key named 'display_n'`

This happens with guake install from:
* `apt` (Ubuntu 18.04, guake 3.0.X)
* `pip` (guake 3.3.2)
* `make` from source code (master branch)

This didn't happen a week ago, and it looks like for some reason gtx decided to look for this `display_n` instead of the `display-n` key that exists.

aichingm:
The main thing in this PR is that context menus now related to the terminal on which they ware called, the other big thing is the movement of code away from guake_app to own classes/files and the utils package.

Summary: 
* Fixed typo in bgcolor to fgcolor in https://github.com/Guake/guake/compare/master...dev/notebook-rewrite#diff-e41ec8d6a3b1c2088836632a5c13ec40L595
* Moved add, remove, and delete page methods from guake_app.py to notebook.py
* Removed context menus from the glade file since the are heavily edited in the code anyways
* (see above) Rewrote the context menus to support per terminal menus (so that the terminal on which was clicked can be passed to the click-handler this comes in handy when there are multiple terminals on one notebook page)
* Remove fullscreen logic from guake_app  and add a FullscreenManager in the utils package, also remove window calculations from guake_app and move it to the utils package
* Remove the dead (returned early without doing anything) proxy code since it was disabled since the gtk3 port and needs to be ported anyways
* I moved all dialogs (PromptQuitDialog, RenameDialog and SaveTerminalDialog) to dialogs.py
* I moved a lot of code (menus, dialogs, customcommands, themes) to its own files
* I moved terminal related code (like the setup process) to the Terminal class
* Remove more dead code (on_resizer_drag)
* Added some comments to indicate where methods are use so that they could be moved to some other place)

I also added a "copy url" context menu entry so that  #1363 can be closed and fixed #1370


releasenotes/notes/add-copy-url-b39441ee986bf333.yaml
```diff
@@ -0,0 +1,2 @@
+features:
+  - add a new option in the context menu (copy url)
```
gsemet: cool. no need to split these into 3 files

gsemet:
That is what I call a very good job ! You're a rock star :)

Merged !

releasenotes/notes/add-copy-url-b39441ee986bf333.yaml
aichingm: Sure, sorry about that! 🙂

PR #1391, opened 2018-08-31
update German translation

aichingm:


aichingm:
Closing, is contained in #1392

PR #1393, opened 2018-09-01
fix #1177

### LINKED ISSUE #1177: Scrolling on tab bar to switch tabs doesn't work anymore
The update to GTK3 seems to have removed it

aichingm:
Allow the user to switch tabs by using the scroll/mouse wheel over the tab/tab-bar.

This fixes #1177


gsemet:
Great thanks !

PR #1396, opened 2018-09-03
fix #1395 reimplement main window hide prevention

### LINKED ISSUE #1395: Context menus hide main window in "hide on focus loss" mode

aichingm:
in the utils package there is now a class called `HidePrevention` which can be used to prevent the main window from hiding.
create a new HidePrevention object like `HidePrevention(window)` 
* `may_hide()` returns True if the window is allowed to hide and false if `prevent()` is called from some where 
* `prevent()`sets a flag on the window object which indicates to `may_hide()`  that the window is **not** allowed to be hidden
* `allow()` sets the flag so that it indicates to `may_hide()`  that the window is allowed to be hidden

### DIFF
--- a/guake/boxes.py
+++ b/guake/boxes.py
@@ -9,10 +9,12 @@ from gi.repository import Gdk
 from gi.repository import Gtk
 from gi.repository import Vte
 
+from guake.callbacks import MenuHideCallback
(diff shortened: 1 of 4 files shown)


gsemet:
great !

PR #1417, opened 2018-09-21
Update split character and update fr

gsemet:
@aichingm Can you have a look at this PL, I change the character of your split menu.


aichingm:
new character looks good!

PR #1426, opened 2018-09-25
TerminalNotebook.get_terminals corrected

Ozzyboshi:
The method get_terminals does not require index parameter.
If placed it causes the get_count_tab() dbus method to crash.

get_tab_terminal dbus interface returns 
```
return len(self.guake.notebook.get_terminals())
```
It's passing no arguments, but it seems that get_terminals requires an index.
The index argument is not used anywhere inside the get_terminals method so i decided to remove it.
Please accept this pull request because after the notebook reworking, my project guake-indicator is in trouble since there is a lot to fix.


gsemet:
Approved :)

Ozzyboshi:
i see the execute_command commit is missing

PR #1427, opened 2018-09-25
Update dbusiface.py

Ozzyboshi:
Fixed select_tab interface method.
Now we are forced to use the gtknotebook object, guake.select_tab() was deleted in previous commits.


gsemet:
Thanks !

Issue #1097, opened 2018-01-15
[gtk3] Close: Thinks process is running


egmontkob:
Preferences -> General -> Prompt on close tab: With process running

Try to close a tab where only the initial shell is running.

The gtk2 version closed it without a prompt, as expected. The gtk3 always prompts.

aichingm:
Just tested it with current HEAD doesn't happen anymore/for me. Can you confirm?

aichingm:
Can't reproduce and from reviewing the code this should work fine. **If not please just reopen :wink:**

egmontkob:
Now it's the other way around, it never warns me (not even if I set "Prompt on close tab: Always" and a process is running.) Right clicking on the tab label and choosing "Close" immediately closes the tab. (@ ebf97fac)

I'm not offered the possibility to reopen.

PR #1435, opened 2018-10-03
Various Fixes

aichingm:
Minor fixes:

1. add python-pipenv as make dependency on archlinux 
1. remove unused code from the TerminalContextMenuCallbacks
1. restore prompt on close tab functionality #1097

### DIFF
--- a/guake/callbacks.py
+++ b/guake/callbacks.py
@@ -46,23 +46,6 @@ class TerminalContextMenuCallbacks():
         # this is not implemented jet
         pass
 
-    def on_new_tab(self, *args):
-        self.notebook.new_page_with_focus(directory=self.terminal.get_current_directory())
-
-    def on_rename_tab(self, *args):
-        page_num = self.notebook.find_page_index_by_terminal(self.terminal)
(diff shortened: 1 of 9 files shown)


guake/callbacks.py
```diff
@@ -46,23 +46,6 @@ def on_find(self):
         # this is not implemented jet
         pass
 
-    def on_new_tab(self, *args):
-        self.notebook.new_page_with_focus(directory=self.terminal.get_current_directory())
```
gsemet: Sure it is not used by tab context menu ?

aichingm: Yes the tab context menu implements the on_new_tab method in https://github.com/Guake/guake/blob/master/guake/boxes.py#L365 there is no extra callback object (TabLabelEventBox passes it self as callback object). Since the tab context menu has only 3 on_* methods I decided to do it directly in TabLabelEventBox, maybe this should change in the future to be cleaner.

gsemet: oh, great. So this was a duplicate code :)

PR #1445, opened 2018-10-11
Add an AppStream XML file

suve:
Sorry for the lack of a release note, but I don't have `pipenv` installed on my system.


gsemet:
Thanks for your contribution. Do you know how we can use it? Is there any command upon install to register it somewhere?

suve:
I just pushed an additional commit that adds installing / uninstalling the AppStream XML file to the Makefile.

gsemet:
thanks

PR #1446, opened 2018-10-13
0.8.x dbus fullscreen toggle

ElijahLynn:
Currently fullscreen keyboard shortcut for me is broken, no matter what I set it to and I can't jump to a non 0.8.x branch right now for GTK+3. I used this base DBUS script from https://github.com/Guake/guake/issues/492#issuecomment-324165795 and made this script https://gist.github.com/ElijahLynn/6a5f6b1ef64e39c40ef3626caacbe95b. The script uses 2 new DBUS methods that I added to Guake in this PR: unfullscreen() and fullscreen_toggle(). 

I put the script in `~/bin/guake_fullscreen_toggle.py` then set Ubuntu 14.04 to a custom keyboard shortcut (ctrl+shift+alt+t) with a value of `python /home/elijah/bin/guake_fullscreen_toggle.py` (can't use the `~` home shortcut) to invoke the DBUS fullscreen toggle script and it works nicely now.

The unfullscreen() isn't needed for the toggle script but I left it in because I found it useful during testing with the `d-feet` program.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -15,7 +15,7 @@
 # Free Software Foundation, Inc., 59 Temple Place - Suite 330,
 # Boston, MA 02111-1307, USA.
 
-AC_INIT([guake], [0.8.11], [http://guake-project.org/])
+AC_INIT([guake], [0.8.13], [http://guake-project.org/])
 
 AC_CONFIG_HEADERS([config.h])
 
--- a/configure.ac
+++ b/configure.ac
@@ -15,7 +15,7 @@
 # Free Software Foundation, Inc., 59 Temple Place - Suite 330,
 # Boston, MA 02111-1307, USA.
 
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -15,7 +15,7 @@
 # Free Software Foundation, Inc., 59 Temple Place - Suite 330,
 # Boston, MA 02111-1307, USA.
 
-AC_INIT([guake], [0.8.11], [http://guake-project.org/])
+AC_INIT([guake], [0.8.13], [http://guake-project.org/])
```
gsemet: can you use 0.8.14 ?
The 0.8.13 is actually a released version !

ElijahLynn:
Sure, done.

ElijahLynn:
Thank you!

PR #1452, opened 2018-10-22
New dbus interface method to execute command on focused splitted term…

Ozzyboshi:
Added execute_command_current_termbox dbus function.
The difference between the old execute_command and the execute_command_current_termbox is that the first execute the command in all terminals tied to the current guake tab, the second one sends the command ONLY to the terminal on focus (useful after splitting terms).
This feature  would allow my guake-indicator to perform more complex macros like, for example, opening a new tab, splitting vertically and execute a command in the first split and another in the second.


gsemet:
no pbl :)

PR #1459, opened 2018-11-12
Update installing.rst

tecty:
A way to successful install guake in fedora 29, not tested in 28.

Just a simple update of documents. Nothing changed in code of guake.


gsemet:
hi. Can you update your pull request?

tecty:
ok, I updated it.

tecty:
There's a bug when using guake command. When I choose 'close when loose focus' and use F12, it will loose focus and then open again.

Issue #1450, opened 2018-10-19
New code for use_mouse doesn't work correctly


michaelarnauts:
On my system (ubuntu 18.10), the code to detect the current screen is not working. If I use the old code from before commit c4b5b202ed5b0df3d7bf81046a40824a5ef2f778 it is working fine.

The issue is that when my cursor is on my primary screen, the `dest_screen` alternates between my primary screen (3), and screen 1 (my primary screen is screen 3, in a setup of 1 3 2). 

Applying this patch on the latest master makes it work again.

```diff
diff --git a/guake/utils.py b/guake/utils.py
index ab1124a..20cf320 100644
--- a/guake/utils.py
+++ b/guake/utils.py
@@ -274,19 +274,11 @@ class RectCalculator():
         dest_screen = settings.general.get_int('display-n')
 
         if use_mouse:
-            """
-            TODO this is ported from widget.get_pointer() to
-            GdkSeat.get_pointer(), but this whole method could be
-            ported to Gdk (eg. gdk_display_get_default_screen(...)
-            and gdk-screen-get-n-monitors(...))
-            """
-            gdk_window = window.get_window()
-            if gdk_window is not None:
-                display = Gdk.Display.get_default()
-                seat = display.get_default_seat()
-                device = seat.get_pointer()
-                win, x, y, _ = gdk_window.get_device_position(device)
-                dest_screen = screen.get_monitor_at_point(x, y)
+
+            # TODO PORT get_pointer is deprecated
+            # https://developer.gnome.org/gtk3/stable/GtkWidget.html#gtk-widget-get-pointer
+            win, x, y, _ = screen.get_root_window().get_pointer()
+            dest_screen = screen.get_monitor_at_point(x, y)
 
         # If Guake is configured to use a screen that is not currently attached,
         # default to 'primary display' option.
```

printfsam:
@michaelarnauts - You're solution worked for me.

Figured I'd put it here for reference.
1. Ran `guake`
2. Error: `AttributeError: 'X11Display' object has no attribute 'get_default_seat'`

(I'm not really versed in gdk at all so my original hack was to just comment out the use_mouse statement which did the trick)

PavelSmokorovskiy:
write please step by step solution to fix the problem with guake wrong monitor position

L3vi47h4N:
This fix worked for me. I performed the following steps on Debian with guake as installed from the debian unstable repo.

```
[...]

PR #1460, opened 2018-11-19
Revert use_mouse code to previous behavior

### LINKED ISSUE #1453: Guake 3.4 doesn't correcly appear on mouse display
- **Guake 3.4.0**
 - Linux Mint 19 (x64)

I have the option "Appear on mouse display" enabled but the window doesn't always appear on the display where the mouse is.

I have 3 monitors - 1 left, 2 center (primary) and 3 right.
If I enable the Guake window while the mouse is on the 2nd (primary) monitor the window will alternatively appear on monitor 1 and monitor 2 (clicking to show/hide the window multiple times). It will appear once on monitor 1, then on monitor 2, then repeat.

If the mouse is on monitor 3 the window will alternatively appear on monitor 1 and 3.

I've recently updated from Guake 3.3.2 where I didn't have this problem.

michaelarnauts:
Revert the use_mouse code to previous behavior. Fixes #1453 and #1450

### DIFF
--- a/guake/utils.py
+++ b/guake/utils.py
@@ -274,19 +274,11 @@ class RectCalculator():
         dest_screen = settings.general.get_int('display-n')
 
         if use_mouse:
-            """
-            TODO this is ported from widget.get_pointer() to
-            GdkSeat.get_pointer(), but this whole method could be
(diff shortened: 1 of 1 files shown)


gsemet:
thanks :)

PR #1465, opened 2018-12-01
Update translations

gsemet:


gsemet:
too much change. must do them differently

PR #1415, opened 2018-09-20
Split terminal

aichingm:
Fixes #71 

- Splitting as deep as you want
- Resize the split area with the mouse
- New context menu entries for splitting and closing terminals

Keybindings for:
- Splitting horizontal and vertical
- Moving the focus to the terminal above, below, left, or right
- Resizing the terminals up, down, left and right
- Closing a terminal

PR #1467, opened 2018-12-05
Each Workspace gets its own set of Tabs

### LINKED ISSUE #32: Workspace-specific guake

freimair:
closes #32 , finally

Inspired by #820, Guake now can provide workspace-specific tab sets.

Made possible by #1415.


freimair:
Now even the last styling check passes :)

PR #1475, opened 2018-12-14
remove duplicate event bind that causes issues with quick_open

David-Solodukhin:
previously, 2 events were firing for every single click which meant that a quick open cmd would be run twice for every one click. e.g. I had 'xdg-open %(file_path)s' for the quick_open command but this would get executed twice for every one ctrl+click.


robert-matusewicz:
I think that is resolved now. I tried to reproduce that Ubuntu 19.10 (Guake 3.6.3) and every time I CTRL+click on a link, it opens only one tab.

PR #1480, opened 2018-12-31
 fix data path in paths.py

MahaGamal:
Issue on Guake Terminal 3.4.0

I tracked it down to `ab7b44a`.
 If I undo this change, guake will start.

### DIFF
--- a/guake/paths.py.in
+++ b/guake/paths.py.in
@@ -35,7 +35,7 @@ def get_default_package_root():
 
 
 def get_data_files_dir():
-    d = os.path.dirname(os.path.dirname(sys.modules["guake"].__file__))
+    d = os.path.dirname(sys.modules["guake"].__file__)
     p = os.path.basename(os.path.abspath(os.path.join(d, "..")))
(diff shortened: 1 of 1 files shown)


gsemet:
How did you installed guake?

MahaGamal:
Using Make or pip3, I had same problem with data path until editing path it’s working fine

aichingm:
How did your paths look before and after the change?

ulidtko:
This isn't a good fix, I give my **Don't merge** here. :-1: 

The proper way is to use `pkg_resources.get_resource_filename()` instead of the hacking around `__file__`. The `paths.py.in` even imports that function — but doesn't use it.

See e.g. [this stackoverflow](https://stackoverflow.com/questions/39104/finding-a-file-in-a-python-module-distribution).

ulidtko:
As a side note: patches like this should be **rebased** onto the latest master, not merge it in.

gsemet:
I agree to not merge that, but I would like to ditch pkg_resource at one point. The real solution is to use importlib_resource library instead.

PR #1481, opened 2019-01-04
Improvments for #1467

aichingm:
- replace show + hide with grab_focus
- add a notebook manager to remove some of the python magic and make code more maintainable
- fix settings only applied to the active workspace if more the 1 is used
- fix prompt quit dialog numbers when more then 1 workspace is used
- fix dbus interface when more then 1 workspace is used
- minor fixes (see commit messages)
- update German translation


gsemet:
thanks loot great!

PR #1493, opened 2019-01-25
Fixes #1492

### LINKED ISSUE #1492: "guake --select-tab" always returns "invalid index"
Switching tabs by command line with `guake --select-tab <index>` works fine but it complains about "invalid index" regardless of the outcome since recently.  Not a huge problem, but it broke my bootstrap script.

I tried to pinpoint the error, but lost track in the realms of gobject. A simple check for the number of tabs instead of relying on gobject return values might be a solution.

Guake: https://github.com/Guake/guake/commit/e0dba674ee819962efc1f27a27417b94c3c67fa2
Cinnamon: 4.0.9-1
gtk3: 3.24.4-1
python-gobject: 3.30.4-1
OS: Arch Linux

bendev:
- select_tab checks for correct index before switching tab
- "invalid index" error message goes to stderr now

### DIFF
--- a/guake/main.py
+++ b/guake/main.py
@@ -340,9 +340,11 @@ def main():
 
     if options.select_tab:
         selected = int(options.select_tab)
-        i = remote_object.select_tab(selected)
-        if i is None:
-            sys.stdout.write('invalid index: %d\n' % selected)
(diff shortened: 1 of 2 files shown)


gsemet:
thanks !

PR #1508, opened 2019-03-29
Implement option for 8-bit bright text

### LINKED ISSUE #1089: FYI: New APIs in VTE 0.52
FYI: VTE 0.52 (to be released in March 2018) adds 3 new functionalities and their corresponding APIs (already present in git). It's up to you whether you want to hook up and make config options for them.

Hopefully Ubuntu Bionic 18.04 LTS is going to ship VTE 0.52. If you care about this distro and would like your app to offer these features in this long term support release, you should probably release a new version of your app in time which (conditionally to this VTE version) enables these features.

- `vte_terminal_{get,set}_cell_{height,width}_scale`: Increased line spacing and character spacing mostly for accessibility for people with imperfect sight. (`gnome-terminal` added two new spinbuttons below the font chooser.)

- `vte_terminal_{get,set}_bold_is_bright`: Whether escape sequences such as `\e[1;35m` switch to bright colors in addition to bold (legacy behavior) or switch to bold and leave the color intact (new behavior, required for solarized, base16 and similar modern color schemes). (`gnome-terminal` added a new checkbox "Show bold text in bright colors" at the bottom of the Colors tab.) (And while at it, we've removed "Allow bold text" from `gnome-terminal`'s UI, and will probably start deprecating its API in about a year. I see no sane reason for this option to exist.)

- `vte_terminal_{get,set}_text_blink_mode`: VTE now implements blinking text (`\e[5m`) and enables it by default, this option lets you disable it or make it conditional to the terminal being (un)focused. (`gnome-terminal` added an "Allow blinking text:" "Never" / "When focused" / "When unfocused" / "Always" dropdown to the General tab of profile prefs.)

Please refer to VTE's source tree for details, e.g. compile with `--enable-gtk-doc` and check the generated docs under `doc/reference/html/index.html`.

### LINKED ISSUE #1507: Guake always uses dark colors
In a recent update, Guake started only using the theme's dark colors (the top row in Guake's appearance settings). Guake always displays the darker color, even if it should display the brighter one.
In other words: `\033[0;32m` and `\033[1;32m` produce the same color. They produced different colors previously. The latter one still causes text to be bold though.

  - **Guake Version**: 3.4.0
  - My System: Arch Linux 64 bit (fully up to date)
  - WM: KDE Plasma

eli-schwartz:
In vte 0.52, a new option was added to emit only bold when bold text is emitted, and made the default behavior in vte 0.56.

The motivation for the change is to allow better 256-bit color in applications and colorschemes. However, traditionally, applications assumed that bold text would also be bright, and many applications look quite bad with this setting. It is therefore useful to allow users to decide whether to utilize this behavior.

Fixes #1507
Closes #1089

### DIFF
--- a/data/org.guake.gschema.xml
+++ b/data/org.guake.gschema.xml
@@ -282,6 +282,11 @@
             <summary>Allow displaying bold font in Guake terminal.</summary>
             <description>When allow_bold is disabled, any text in Guake terminal intended to print as bold text will be instead rendered as normal text.</description>
         </key>
+        <key name="bold-is-bright" type="b">
+            <default>false</default>
+            <summary>Use bright colors for bold text</summary>
+            <description>If true, text from the first 8 colors that is formatted as bold will also be switched to its bright variant.</description>
+        </key>
     </schema>
     <schema id="guake.style.background" path="/apps/guake/style/background/">
         <key name="transparency" type="i">
--- a/data/prefs.glade
+++ b/data/prefs.glade
@@ -1677,6 +1677,22 @@
                                     <property name="top_attach">2</property>
                                   </packing>
                                 </child>
+                                <child>
+                                  <object class="GtkCheckButton" id="bold_is_bright">
+                                    <property name="label" translatable="yes">Bold text is also bright</property>
+                                    <property name="visible">True</property>
+                                    <property name="can_focus">True</property>
+                                    <property name="receives_default">False</property>
+                                    <property name="halign">start</property>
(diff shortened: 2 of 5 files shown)


egmontkob:
I made an inline comment, but I'm not familiar with this part of github, not sure if a notification went out or what. So to be safe, here's it again:

You should check for 0.52 – the feature is available since then, the flip of the default in 0.56 is irrelevant if you add this option. Thanks! :)

Issue #114, opened 2013-04-30
Feature Request: Save Sessions


kikecalpe:
Feature Request: Save Sessions  http://guake.org/ticket/367
Reported by:    huahaiy     Owned by:   somebody
Priority:   major   Milestone:  0.5.0
Component:  guake   Version:  
Keywords:       Cc:  
Description

I use guake at work as my main terminal and use KDE at home. One feature from konsole that I miss in guake is being able to save the session state, so that all my previously opened and named tabs are restored when I restart guake.

cirosantilli:
+1. I'd also like to automatically save and reopen the seesion, as asked at: http://askubuntu.com/questions/98754/maintain-terminal-session-after-reboot

Suor:
+1 would be very nice to have.

MbarekFiras:
+1

calvinwankhede:
+1

Phiwa:
+1

byted:
+1

kodedyret:
+1

puregardenia:
I install the application by compile the source file, and It can't get the status and config before restart the system. most important, It can't get the CONFIG! when I restart the system, It like a new application just now. Anybody know the problems

puregardenia:
I got the solution [#788](https://github.com/Guake/guake/issues/788)

DareDevil73:
+1

zasca:
+1

avrilfanomar:
+1

kecalcze:
+1

fabriciogpp:
+1

linxon:
+1

gaursagar:
+1

FedeG:
+1

knochenhans:
+1

avrilfanomarremit:
I've got a solution: I didn't reboot for a month. Works perfectly ;)

AndrewEastwood:
LOL

arturespl:
+1

imranashraf:
+1

frederiksnijder:
+10

kusha:
+1

frederik-labbe:
+9000

curtisz:
@gsemet With this last **0.8.11** milestone removed, what's the plan for this feature request?

gsemet:
I need this feature myself, so... patience ! :)

nbozeman:
Let's do it! +1

rsansores:
[...]

PR #1515, opened 2019-04-12
Add save/restore tabs function

mlouielu:
Hi all, I make a simple version of save/restore tabs function into guake

You can test it with right click on the notebook context menu, it will show up "Save Tabs" & "Restore Tabs" menu item for you to use it.

Since I'm not that familiar to gtk and guake code, there has some issue need to resolved:

1. Currently, it will only save/restore pages directory and its label text, inside the `TerminalBox`, the split will not be saved, should we deal with it? (since I'm not really sure on this)
2. It should add hotkey (accel-key?) for save-tabs and restore-tabs
3. Did guake have some config folder in `.config`?  Currently, the tabs config file default output will be `.guake_tabs.json`, it definitely have some better place and name for it.
4. Cont. previous, the tabs config file output name should be configurable? (again, I'm not familiar with the gtk config stuff :((  )

thanks.

ref: #114


gsemet:
Yes, that looks cool :) I like the use of ~/.config/guake/session.json.
Accelerator would be cool but this can wait.

Does it restore automatically upon start ?

mlouielu:
So for the tabs config file, it will need to ensure within `XDG_CONFIG_HOME`, maybe a new dependency on `pyxdg` to deal with it

Restore automatically upon start can be made, but I'll prefer to let user choice (just like the browser), and probably auto-save (when changing name / add tab ..etc)

Maybe we can use the desktop notification just like when the guake startup?

And last, for the security issue, did the label text and directory path have been sanitized when accepting the input?

gsemet:
That’s fine for me for the new dep.
You can add a notification that’s a good idea.

There is no sanitation of the text, sadly.

gsemet:
Look great to me. Let’s merge

mlouielu:
@gsemet 6 years issue done! Hooray!

PR #1516, opened 2019-04-13
Enhance restore tabs (with prefs)

mlouielu:
Cont. #1515 

Add 3 prefs for save/restore tabs

* restore-tabs-startup: when enabled, it will restore tabs when startup
* restore-tabs-notify: when enabled, it will notify user after tabs restored (except startup)
* save-tabs-when-changed: when enabled, it will automatically save tabs session when changed (new/del/reorder)


gsemet:
thanks a lot !

PR #1521, opened 2019-04-13
Add about dialog unittest

mlouielu:
@gsemet are you familiar with PyGObject unittest? I'm not sure if this a correct way or not, maybe somewhere need to mock up. Otherwise window will popup very quickly...

### DIFF
--- a/guake/tests/test_about.py
+++ b/guake/tests/test_about.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+
+import pytest
+
+from guake import guake_version
+from guake.about import AboutDialog
+from locale import gettext as _
+
+
+class TestAboutDialog:
+
+    @classmethod
+    def setup_class(cls):
+        cls.ad = AboutDialog()
+        cls.dialog = cls.ad.get_widget('aboutdialog')
+
+    @classmethod
+    def teardown_class(cls):
+        cls.dialog.destroy()
+
+    @pytest.mark.dialog
+    def test_version_text(self):
+        assert self.dialog.get_version() == guake_version()
+
+    @pytest.mark.dialog
+    def test_title(self):
+        assert self.dialog.get_title() == _('About Guake')
--- a/guake/tests/test_about.py
+++ b/guake/tests/test_about.py
@@ -7,21 +7,16 @@ from guake.about import AboutDialog
 from locale import gettext as _
 
 
-class TestAboutDialog:
+@pytest.fixture
+def dialog(mocker):
+    mocker.patch('guake.simplegladeapp.Gtk.Widget.show_all')
+    ad = AboutDialog()
+    return ad
 
-    @classmethod
-    def setup_class(cls):
-        cls.ad = AboutDialog()
-        cls.dialog = cls.ad.get_widget('aboutdialog')
 
-    @classmethod
-    def teardown_class(cls):
-        cls.dialog.destroy()
+def test_version_test(dialog):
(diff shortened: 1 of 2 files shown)


guake/simplegladeapp.py
```diff
@@ -30,7 +30,7 @@
 
 class SimpleGladeApp():
 
-    def __init__(self, path, root=None, domain=None, **kwargs):
```
gsemet: Hehe, you should not taint your class for unit test.
You need to use mock and patch for that

guake/tests/test_about.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+
+import pytest
+
+from guake import guake_version
+from guake.about import AboutDialog
+from locale import gettext as _
+
+
+class TestAboutDialog:
```
gsemet: Prefer the test_* method of unit test with pytest, so you can encapuslate pytest.fixture

guake/tests/test_about.py
```diff
@@ -0,0 +1,27 @@
...
+
+    @classmethod
+    def setup_class(cls):
+        cls.ad = AboutDialog(mock=True)
```
gsemet: So here you need to « mock.patch » the class and not ask it to create the mock for you.

Prefer using « mocker » package (pytest_mock package) it is way more powerful

gsemet:
Hello. Great idea to start unit testing guake. I never had the courage to do it. I can help you, i write pytest unit test everyday, but never on pygobjec sadly.

Look at my other python packages (cfgtree,..) to see how unit test can be written. Prefer avoid Unittest class and straight « def test_something » function in your test module, along with fixture you will see how powerful it is

guake/tests/test_notebook.py
```diff
@@ -0,0 +1,62 @@
+# -*- coding: utf-8 -*-
+
+import pytest
+import mock
+from mock import Mock, MagicMock
```
gsemet: Prefer mocker

guake/tests/test_notebook.py
```diff
@@ -0,0 +1,62 @@
...
+
+import guake.notebook
+
+
+guake.notebook.TerminalBox = Mock()
```
gsemet: This is not how to mock. You need to use mocker.patch or mock.patch to ensure your unit test restore the initial state at the end of the execution.
I would do something like:
@pytest.fixture(scope=module)
def fake_notebook(mocker):
   mocker.patch(....)

def test_something(fake_notebook):
   ...

mlouielu: OK, I definitely need to first take a look about pytest documentation 🤔

gsemet: There some good article:
- https://codefellows.github.io/sea-python-401d7/lectures/mock.html
- https://flowfx.de/blog/using-the-pytest-mock-plugin/

You might also need to understand how the [conftest.py trick works](https://docs.pytest.org/en/2.7.3/plugins.html?highlight=re) to refactorize fixture.

Thanks a lot for your series of recent contribution, it will help improving guake for everyone to use!

mlouielu:
@gsemet I think the latest version addresses your points :)

guake/tests/test_notebook.py
mlouielu: Just got a timeslot to contribute to the tool that I used daily :)

gsemet:
yes i looks great now :)

PR #1524, opened 2019-04-15
Fix travis-ci with unittest

mlouielu:
* Update to xenial in travis-ci
* Use pygobject / pycairo inside virtualenv, not site-packages (#1525)
* Fix pylint W0621 for unittest fixtures

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,6 @@ install:
 
 before_script:
     - export DISPLAY=:99.0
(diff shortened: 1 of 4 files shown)


mlouielu:
issue: 3.6 run in pipenv with 3.7m...?

```
Using /opt/pyenv/versions/3.7/bin/python3.7m (3.7.1) to create virtualenv…
```

PR #1529, opened 2019-04-16
Fix CLI with 2 toggle to hide when combine some options

### LINKED ISSUE #1499: Show/hide key confused with `guake --show -s 0`
**a bug**: after firing up Guake with `guake --show -s 0` the toggle key( default F12) is confused as it doesn't hide the guake window and second press is needed.  This is weird a bug as it doesn't happen, ie. works fine, when using `guake --show` without the argument for tab swiching.
 
  - **Guake Version**: 3.4.0
  - Fedora 29

mlouielu:
This fix #1499

### DIFF
--- a/guake/main.py
+++ b/guake/main.py
@@ -375,11 +375,11 @@ def main():
 
     if options.show_preferences:
         remote_object.show_prefs()
-        only_show_hide = False
+        only_show_hide = options.show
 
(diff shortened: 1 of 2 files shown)


mlouielu:
@gsemet I think the releasenote of this PR is invalid yaml :(

PR #1533, opened 2019-04-17
Search in terminal

mlouielu:
For #116, currently we have two ways to implement it, one is the search dialog (we have search.glade), another is like gedit or iterm using a small search box to search the terminal.

This PR trying to implement the latter, this is how it work to float the search box on the page:

* RootTerminalBox: Gtk.Box -> Gtk.Overlay
* RootTerminalBox.child (TerminalBox/DualTerminalBox) -> `add` into overlay
* Search box -> `add_overlay` into overlay

In this way, we can then implement the search function base on `vte.Terminal`.

Some problem:
- [x] Not test in different scenario
- [x] Search box will not hide in focus-out-event (this is tricky... we cannot hide the box and show it out, it will let the search fall back to first item)
- [x] Next/Prev button didn't set sensitive correctly
- [x] It will only search in last focus terminal (if you split, it will not search in all terminal in same RootTerminalBox)

VTE problem:
- [ ] It cannot highlight all match item (long time feature request)

TODO:

- [x] Prefs setting
- [x] Prev search with SHIFT
- [x] Cleanup RootTerminalBox other related method (from Gtk.Box to Gtk.Overlay) and check if somewhere breakup by this

Feel free to test it, you can test it with `CTRL+SHIFT+F` to search

### DIFF
--- a/data/org.guake.gschema.xml
+++ b/data/org.guake.gschema.xml
@@ -328,6 +328,11 @@
         </key>
     </schema>
     <schema id="guake.keybindings.local" path="/apps/guake/keybindings/local/">
+        <key name="search-terminal" type="s">
+            <default>'&lt;Control&gt;&lt;Shift&gt;f'</default>
+            <summary>Search terminal</summary>
+            <description>Show search box to search text inside the terminal</description>
+        </key>
         <key name="quit" type="s">
             <default>'&lt;Control&gt;&lt;Shift&gt;q'</default>
             <summary>Quit</summary>
--- a/guake/boxes.py
+++ b/guake/boxes.py
@@ -9,6 +9,7 @@ from gi.repository import Gdk
 from gi.repository import Gio
 from gi.repository import Gtk
 from gi.repository import Vte
+from gi.repository import GLib
 
 from guake.callbacks import MenuHideCallback
 from guake.callbacks import TerminalContextMenuCallbacks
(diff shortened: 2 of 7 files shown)


gsemet:
Great ! I agree the search box is not ideal, subpanel a-la iterm2 or Firefox is much better ! Can’t wait to test it !
Don’t forget to have « reverse direction » by default.
The other problems can be accepted as long as the search works. For the dependency on a vte patch, I know it is a long time request, maybe one day it will be implemented !

Thanks for your effort, I really appreciate :)

mlouielu:
@gsemet I think it is good for merge now, please take a test on it :)

gsemet:
yep! i'm testing it now!

austinmw:
@gsemet Hi, I'm able to see the search box with Ctrl+shift+F, but it doesn't seem to actually work for me. It doesn't highlight or jump to any text. I tried a few different themes in case the color was hard to see.

jrao:
Same as @austinmw for me on Ubuntu 20.04, Gnome desktop.

```
$ guake --version
Guake Terminal: 3.6.3
VTE: 0.60.3
VTE runtime: 0.60.3
Gtk: 3.24.20
```

Axel29:
Same for me, I have the search box but nothing gets searched ^^"

pevogam:
Any opened issue or pull request regarding the unusable search box? One can type inside but nothing gets highlighted or done once they type enter, or "next" or "previous" arrow result.

gsemet:
Hello. Feel free to work on it and provide this pull request !

PR #1556, opened 2019-04-24
Add guake_app unittest

mlouielu:
Add unittest for `guake_app`

### DIFF
--- a/guake/tests/test_guake.py
+++ b/guake/tests/test_guake.py
@@ -1,2 +1,171 @@
-def test_000():
-    pass
+# -*- coding: utf-8 -*-
+# pylint: disable=redefined-outer-name
+
+import json
+import os
+import tempfile
+import time
+
+from pathlib import Path
+
+import pytest
+
+import guake.guake_app
+
+from guake.common import pixmapfile
+from guake.guake_app import Guake
+
+
+@pytest.fixture
+def g(mocker, fs):
+    mocker.patch('guake.guake_app.Guake.get_xdg_config_directory', return_value=Path('/foobar'))
+    mocker.patch('guake.guake_app.shutil.copy', create=True)
+    mocker.patch('guake.guake_app.notifier.showMessage', create=True)
+    mocker.patch('guake.guake_app.traceback.print_exc', create=True)
+    fs.pause()
+    g = Guake()
+    fs.add_real_file(pixmapfile('guake-notification.png'))
+    fs.resume()
+    return g
+
+
+# Accel Test
+
+
+def test_accel_search_terminal(g):
+    nb = g.get_notebook()
+    page = nb.get_nth_page(0)
+    assert not page.search_revealer.get_reveal_child()
+
+    g.accel_search_terminal()
+    assert page.search_revealer.get_reveal_child()
+
+
+def test_accel_search_terminal_debounce(g):
+    nb = g.get_notebook()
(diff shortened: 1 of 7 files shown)


guake/tests/test_guake.py
```diff
@@ -1,2 +1,106 @@
-def test_000():
-    pass
+# -*- coding: utf-8 -*-
+# pylint: disable=redefined-outer-name
+
+import json
+import os
+import pathlib
```
gsemet: I recommend using [path.py](https://pypi.org/project/path.py/) instead of pathlib :)

guake/tests/test_guake.py
```diff
@@ -1,2 +1,106 @@
...
+
+import guake.guake_app
+from guake.guake_app import Guake
+
+
+@pytest.fixture
+def g(mocker):
+    mocker.patch(
+        'guake.guake_app.Guake.get_xdg_config_directory',
+        return_value=pathlib.Path(tempfile.gettempdir())
```
gsemet: Better, you can use [fakefs](https://pypi.org/project/pyfakefs/)

guake/tests/test_guake.py
```diff
@@ -1,2 +1,106 @@
...
+    nb = g.get_notebook()
+    assert nb.get_n_pages() == 3
+    assert nb.get_tab_text_index(0) == '1'
+    assert nb.get_tab_text_index(1) == '2'
+
+    nb = g.notebook_manager.get_notebook(1)
+    assert nb.get_n_pages() == 1
+    assert nb.get_tab_text_index(0) == '4'
+
+    d1.cleanup()
```
gsemet: I would try catch this to be sure to alway call cleanup even if the test breaks. Again, with fakefs, no need :)

mlouielu:
hmm, just break the python 3.5 unittest...

guake/tests/test_guake.py
mlouielu: I still stuck with `pathlib.Path`, since most of the `Path` object in the code is using `pathlib`, and using `path.py` make some breakage, e.g. `Path().exists()` (pathlib) -> `os.path.exists(Path())` (path.py), `with Path().open() as f` (pathlib) -> `with open(Path()) as f` (path.py)

guake/tests/test_guake.py
mlouielu: I still stuck with `pathlib.Path`, since most of the `Path` object in the code is using `pathlib`, and using `path.py` make some breakage, e.g.
* `Path().exists()` (pathlib) -> `os.path.exists(Path())` (path.py)
*  `with Path().open() as f` (pathlib) -> `with open(Path()) as f` (path.py)

guake/tests/test_guake.py
mlouielu: Address it with pyfakefs

guake/tests/test_guake.py
gsemet: ok for me.

But path.py still provides the .exists() function, .open() like pathlib does. It is basically pathlib + mkdirs + convert to str smoothly :)

gsemet:
thanks that's great !

guake/tests/test_guake.py
mlouielu: hmmm, so it is caused by fakefs? I'm not sure about this :(

PR #1557, opened 2019-04-26
Make `make` with different python interpreter

mlouielu:
User can choice which python interpreter when using `make`, e.g.:

    $ PYTHON_INTERPRETER=python3.5 make dev


gsemet:
damned makefile syntax, forgot the ? in the ?=

PR #1563, opened 2019-04-26
Fix search box background color to follow theme

### LINKED ISSUE #1561: New feature "search box" doesn't look good with dark GTK themes.
Hi, i just tested this new feature and i must say search box doesn't look good with dark theme because overlay background is light-colored and dark themes usually have light-colored fonts, so it's barely visible and it's very hard to read what you type. At first i thought it's not working because i couldn't see the text at all until i really focused my eyes on the search box. Also, background is not transparent as the rest of the terminal, so it looks really ugly and it really stands out from the rest of design. At least make background the right color (the one provided by the current selected GTK theme).

  - **Guake Version**: 3.6.0
  - System: Solus Budgie
    - i packaged Guake myself to test new 3.6.0 features, this version is still not in official Solus package repository

This is probably due to hardcoded value of background color as can be seen here:
https://github.com/Guake/guake/blob/0bba1b06e63584193157e5f670f67e7aa5b17a52/guake/boxes.py#L118-L121

mlouielu:
Fix #1561

### DIFF
--- a/guake/boxes.py
+++ b/guake/boxes.py
@@ -118,11 +118,12 @@ class RootTerminalBox(Gtk.Overlay, TerminalHolder):
 
         # Frame
         self.search_frame.set_margin_end(12)
+        self.search_frame.get_style_context().add_class('background')
         css_provider = Gtk.CssProvider()
         css_provider.load_from_data(
             b'#search-frame border {'
             b'    padding: 5px 5px 5px 5px;'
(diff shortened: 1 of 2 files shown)


gsemet:
Indeed this is a better solution. Can you add a reno note please so we can release 3.6.1?

mlouielu:
Added reno

PR #1565, opened 2019-04-29
Add CSS styling and transition for notebook tabs

mlouielu:
Inspired by Firefox, styling TerminalNotebook.tabs:

* Set tab left/right margin to 0px
  (original is 4px each, cause 8px gap between tabs)

* Set tab border, and its transition when hover, checked


mlouielu:
Before:
Selection_970
After:
Selection_971

PR #1566, opened 2019-04-29
Update translations, revert #1565 and fix site-packages

gsemet:
@mlouielu  I had to revert your patch, here is how it look on adwaita theme:
Capture du 2019-04-29 10-31-41

I guess we should not play too much with theme. Just adjust your padding or so, not other stuff


mlouielu:
Adwaita-dark should work, too:

Selection_972

Adwaita-dark [has defined](https://github.com/GNOME/gtk/blob/f27ecde1e28c9aef7fe3b6aca2ac9c519d300e51/gtk/theme/Adwaita/gtk-contained-dark.css#L913) the color on it

mlouielu:
Also, in #1565, the stylesheet didn't change with `box-shadow` color, it shouldn't affect `tab:checked` on it

gsemet:
I understand, but reverting to this commit fix the theme weirdness

mlouielu:
@gsemet Are you using HighContrastInverse instead?

gsemet:
nop. What is it? I use the dark theme indeed

mlouielu:
@gsemet You are right, after testing several theme, I gave up :(

They use different method to implement the `:checked` tab..

gsemet:
adwaita also has this 4px padding, but if we change something on the theme, we should try it with as many theme as possible, to avoid sideeffect

Issue #708, opened 2016-02-11
Drag & Drop a text from another application


satriani-vai:
Can't copy a text part from eg. Gedit to Guake per drag & drop.
That would be great if it would work.
Thanks.

Stibbons:
Planned for guake 3 :) soon

satriani-vai:
Thanks, we are looking forward ;)

Stibbons:
If you want this feature, please vote on [FeatHub](https://feathub.com/Guake/guake)

PR #1570, opened 2019-04-30
Add drag-n-drop for terminal (text & uris)

mlouielu:
Implement #708, dnd support UTF-8 & locale charset.

([FeatHub: Drag and drop text](https://feathub.com/Guake/guake/+35))


gsemet:
nice !

PR #1575, opened 2019-05-02
Avoid spurious resize event when showing fullscreened window

eMPee584:
In ncurses apps that handle SIGWINCH, a resize event is happening when guake shows in fullscreen mode.
The call to `self.window.show_all()` was added in 4f4ab35d5cb2363dd464e6b252d1c798bd01c2fb (2007-10-06) and when it might even be needed is not quite clear.

### DIFF
--- a/guake/guake_app.py
+++ b/guake/guake_app.py
@@ -581,7 +581,8 @@ class Guake(SimpleGladeApp):
             self.add_tab()
 
         self.window.set_keep_below(False)
-        self.window.show_all()
+        if not FullscreenManager(self.settings, self.window).is_fullscreen():
+            self.window.show_all()
         # this is needed because self.window.show_all() results in showing every
         # thing which includes the scrollbar too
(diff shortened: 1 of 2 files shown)


gsemet:
Hi. I don’t really the issue with having this event, if the window change size (you pass from not full screen to full screen)

eMPee584:
Probably not when toggling between windowed and fullscreen mode, but when showing fullscreened guake after it was hidden.
-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

eMPee584:
taste case:
1. run f.e.  `ranger` with `guake` in fullscreen mode
2. hide/show `guake` and watch the curses app being resized

eMPee584:
(uhm just for the record)
… found out that that resizing happens only when the scrollbar is disabled (sorry I didn't mention this diversion from the default setup)... it looks like the area would be reserved for a scrollbar on the right side, but that instantly be hidden again.

PR #1577, opened 2019-05-03
Add install/uninstall-dev-locale to ensure dev-locale works

mlouielu:
`guake/paths.py` in dev return `get_default_locale_dir()`, which return `guake/guake/po`.

Add install/uninstall-dev-locale to support locale in dev.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,8 @@ reset:
 
 all: clean dev style checks dists test docs
 
-dev: clean-ln-venv ensure-pip pipenv-install-dev requirements ln-venv setup-githook prepare-install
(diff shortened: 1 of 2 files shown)


gsemet:
I would prefer to fix paths.py to return “po” instead of copying to guake/po

mlouielu:
So it should be separate to two steps:
1. Fix paths.py return with `po/locale`
2. Install `*.mo` into po/locale/*/LC_MESSAGES/guake.mo`

gsemet:
I am looking to remove the mess around paths.py and use a sub module of the guake directory for the data directory maybe this will be applicable for locales as well. In dev things should not be so different.

Issue #951, opened 2017-05-14
Multi-Display positioning incorrect


orbenjc:
When setting the alignment of the window to "right" on display 2, the terminal appears in the middle of display 2. On the "center" alignment the window appears on the left of display 2 and "left" positions the terminal on the right of display 1. The monitors being used are 2 1920 x 1080 monitors. The operating system is Ubuntu 16.04.2 LTS.

orbenjc:
Update:
I have updated from the current Ubuntu Software Center version to version 0.8.8.1 and a majority of the issue is resolved. The terminal frame now appears in the correct area relatively. The key issue now is that there is about a 30 pixel gap between the right side of the terminal and the right side of the monitor. I am currently attempting to find the settings for positioning the frame and edit them.

orbenjc:
Update 2:
This issue appears to be caused by the scroll bar. In my current Ubuntu 16.04.2 distribution the scroll bard floats and only appears when the mouse is hanging over it, there is always a gap there. Hiding the scroll bar leaves the gap there. I tested in a slightly modified version of GNOME where the scroll bar does not float, there is no gap on the modified system.

Issue #607, opened 2015-09-08
Guake doesn't show correctly on multiple monitors


peperunas:
I use 1 external monitor. I tried to hack the source code but without results.
The screenshots shown here, clearly illustrate the bug.

OS: Gentoo
DE: Gnome 3.16.2
Guake: 0.8.0

---

**Monitors' disposition:**

![schermata da 2015-09-08 10-32-24](https://cloud.githubusercontent.com/assets/6033387/9730215/46f03af0-5615-11e5-827c-8a4e051bf381.png)

---

**Fullscreen:**

**Configuration:**
![schermata da 2015-09-08 10-31-24](https://cloud.githubusercontent.com/assets/6033387/9730201/293581e6-5615-11e5-9420-0cfd034bf533.png)

**Screens' view:**

_Guake open on monitor 1 (here the behaviour is correct):_
![schermata da 2015-09-08 10-43-08](https://cloud.githubusercontent.com/assets/6033387/9730381/72cdc47a-5616-11e5-86cd-ef169bee63ee.png)

_Guake open on monitor 2 (doesn't work here, it doesn't even show on the other side):_
![schermata da 2015-09-08 10-43-08](https://cloud.githubusercontent.com/assets/6033387/9730383/73c66274-5616-11e5-8714-578768e6b598.png)

---

**Using custom width while being centered:**

**Configuration:**

![schermata da 2015-09-08 10-37-00](https://cloud.githubusercontent.com/assets/6033387/9730298/e0d54b7e-5615-11e5-8cba-3bfd1533e8b1.png)

**Screens' view:**

_Guake open on monitor 1 (it doesn't appear on the selected monitor, instead it is incorrectly placed on the other monitor):_
![schermata da 2015-09-08 10-42-03](https://cloud.githubusercontent.com/assets/6033387/9730358/42c3b73a-5616-11e5-96d7-74670b9b3985.png)

_Guake open on monitor 2 (here, the behaviour is correct):_
![schermata da 2015-09-08 10-42-21](https://cloud.githubusercontent.com/assets/6033387/9730362/4e898fae-5616-11e5-8413-43828634d747.png)

Stibbons:
yes this is a bad designed code that need to be completely reworked... I don't have this setup here so it is hard for me to reproduce it, but there is still some bugs in the window placement. And Gtk2 doesn't help us a lot

MalikKeio:
I had exactly the same problem on my Ubuntu box.
[...]

PR #1579, opened 2019-05-03
Fix window positioning

michaelarnauts:
This MR changes the code that positions the window. The biggest issue was that the width of the window wasn't accounting for an offset. This made it really hard to make guake use the full width if you have a dock on the side.

I've also removed the unity code since that isn't working anymore, and can now be easy configured by specifying the width of the dock.

Also, there was a duplicate call to `window.move`.

This makes it work for me, but will probably need more testing.

References #565 #607 #951

### DIFF
--- a/guake/utils.py
+++ b/guake/utils.py
@@ -177,11 +177,9 @@ class RectCalculator():
         """
         # fetch settings
         height_percents = settings.general.get_int('window-height')
-
         width_percents = settings.general.get_int('window-width')
         halignment = settings.general.get_int('window-halignment')
         valignment = settings.general.get_int('window-valignment')
-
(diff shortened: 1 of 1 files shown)


mlouielu:
Can it be 2 commits or 2 PR, one for fix window positioning, another for removing unity dead code, that will be more clear for others that view the commit log in the future.

Another thing is, make sure running `make check` to check if the code style is correct or not, we can saw that [CI](https://travis-ci.org/Guake/guake/jobs/527676687#L3884) complain about this :)

gsemet:
Thanks for your PR and thanks for your early review, mlouielu. Nothing more to add, if you can split into two PR and run “make sct” that would be awesome !

gsemet:
Build seems failing

PR #1592, opened 2019-05-13
Update issue templates

mlouielu:
Update to latest GitHub issue templates that can provide the different template to different issue.

Also, update bug report template to mention `guake --support` and make sure user update to latest Guake before filing an issue.


gsemet:
good idea. Thanks

PR #1593, opened 2019-05-13
update translation

gsemet:


mlouielu:
@gsemet I'm not sure how this works, but will the difference between zh_TW / zh_Hant causing the script didn't catch it update on Weblate? Since Weblate still using zh_Hant, and po using zh_TW

gsemet:
Look ok, even if it is named zh_Hant in weblate, it [works with the zh_TW.po file](https://hosted.weblate.org/projects/guake/guake/zh_Hant/#information):
guak-tw

mlouielu:
Ah, I see.

PR #1597, opened 2019-05-15
Fix Guake cannot transparent after composited changed

### LINKED ISSUE #1523: Force transparency on
Guake version 3.4.0 and 3.5.0

Due to the way stuff works and stuff, Guake can sometimes start up before my composite manager, which is not part of my WM. As such, it gets stuck without transparency. I'd like the ability to force transparency to be used even when it's not detected so this issue doesn't happen.

mlouielu:
Fix #1523

It is possible for user to start its display without compositor, then start the compositor after it.
(e.g. start LXDE (w/o compositor), then start xcompmgr. start awesome (w/o compositor), then start compton)

In this scenario, Gtk.Widget / Gdk.Screen will emit a `composited-changed` signal, will indicate the compositor
has changed, we can then update window visual for it.

It is important to mention that `set_visual` should unrealize the window and realize it again, since it should
only call before window is realized.
(ref: https://lazka.github.io/pgi-docs/index.html#Gtk-3.0/classes/Widget.html#Gtk.Widget.set_visual)

### DIFF
--- a/guake/guake_app.py
+++ b/guake/guake_app.py
@@ -196,15 +196,8 @@ class Guake(SimpleGladeApp):
         self.set_tab_position()
 
         # check and set ARGB for real transparency
-        screen = self.window.get_screen()
-        visual = screen.get_rgba_visual()
-        if visual and screen.is_composited():
-            self.window.set_visual(visual)
-            self.window.transparency = True
-        else:
-            log.warn('System doesn\'t support transparency')
-            self.window.transparency = False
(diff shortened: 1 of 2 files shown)


gsemet:
damned, can you rebase?

gsemet:
@mlouielu I have promoted you as developer in Guake, so you can push directly your dev branch in Guake instead of your fork.

mlouielu:
Cool, I may also triage the issue now!

gsemet:
sure

PR #1598, opened 2019-05-15
Remove window draw callback

mlouielu:
It starts from commit: c7e3e8763, but it is no need on these days.

Guake transparent was done by 2 steps:

1. self.window.set_visual
2. Vte.Terminal.set_colors

We didn't paint/draw any other thing on Guake window, so it is
no need to add the draw callback to draw it on SOURCE. Plus, it will
cause some issue on popover (which will draw the square background below
the popover, can be reproduce on weston, and some old ubuntu 16.04)

NOTE: we may need this in future if we want to implement transparent
      scrollbar like tilix.

### DIFF
--- a/guake/guake_app.py
+++ b/guake/guake_app.py
@@ -196,21 +196,8 @@ class Guake(SimpleGladeApp):
         self.set_tab_position()
 
         # check and set ARGB for real transparency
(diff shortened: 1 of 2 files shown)


gsemet:
thanks :) I am not fan of the transparent scrollbar, not so much a priority for me. I would prefer to work on usability (ex: move splitted terminal with the mouse), readability (add small title bar in splitter terminal).

Issue #1603, opened 2019-05-16
Add background image to guake


mlouielu:
FeatHub: https://feathub.com/Guake/guake/+22

We've the reference from https://github.com/gnunn1/tilix/issues/576 & https://github.com/gnunn1/tilix/commit/7ccb7f1c329ec7bb731f13d1f5ab5b157c2032f9 that background image can add back without modifying VTE, only done by cairo.

Here is the WIP preview on guake: 
![Guake!_978](https://user-images.githubusercontent.com/28770408/57824801-e32f3180-77ce-11e9-9ed8-60efe56f2cce.png)

egmontkob:
FYI: See also [vte bug 767575](https://bugzilla.gnome.org/show_bug.cgi?id=767575) and its result `vte_terminal_set_clear_background()` available since vte 0.52.

mlouielu:
Ah, finally know why `vte_terminal_set_background_operator` [was removed](https://gitlab.gnome.org/GNOME/vte/commit/bc532dc160406aaea35756e9a9e8f755c82fddb2).

But I wonder if that is needed? I've use `set_clear_background(False)` when I'm testing with VTE, but for guake, it only get wierd background color whenever I choice which palette:

![Guake!_979](https://user-images.githubusercontent.com/28770408/57838851-13d79100-77f8-11e9-9596-2327e7605234.png)

My solution is to add the `draw` callback with `connect_after`:

https://github.com/Guake/guake/blob/74645dc816b8f46f8ecce12a88b421cb4e3f00b2/guake/notebook.py#L338-L340

Which were use `connect` by [xfce4-terminal](https://git.xfce.org/apps/xfce4-terminal/tree/terminal/terminal-screen.c?id=f784f668abe9dc8a35fa53a2032eddb52ac526a9#n272) and [tilix](https://github.com/gnunn1/tilix/blob/master/source/gx/tilix/session.d#L1002)

So in the lifecycle, it will call the callback after VTE draw:

```
INFO     Showing the terminal
vte_terminal_get_preferred_height()
vte_terminal_get_preferred_width()
vte_terminal_realize()
vte_terminal_map()
vte_terminal_size_allocate()
vte_terminal_get_preferred_height()
vte_terminal_get_preferred_width()
vte_terminal_draw()
draw root_terminal_box
```

Thus, we will not overlapped by VTE draw now. But maybe some performance drop, since we draw the child twice.

[...]

PR #1604, opened 2019-05-16
Add background image to guake

mlouielu:
Impl #1603.

Impl detail please refer to `guake/utils.py:BackgroundImageManager.draw`

### DIFF
--- a/guake/data/org.guake.gschema.xml
+++ b/guake/data/org.guake.gschema.xml
@@ -172,15 +172,15 @@
             <description>Place Guake at: 0: top, 1: bottom</description>
         </key>
         <key name="background-image-file" type="s">
-          <default>''</default>
-		  <summary>Background image filename.</summary>
-		  <description>When setup, the selected background image will show on guake</description>
+            <default>''</default>
+            <summary>Background image filename.</summary>            
+            <description>When setup, the selected background image will show on guake</description>
         </key>
-		<key name="background-image-layout-mode" type="i">
-		  <default>0</default>
-		  <summary>Background image layout mode.</summary>
-		  <description>To select which layout mode want to use for background image.</description>
-		</key>
+        <key name="background-image-layout-mode" type="i">
+            <default>0</default>
+            <summary>Background image layout mode.</summary>
+            <description>To select which layout mode want to use for background image.</description>
+        </key>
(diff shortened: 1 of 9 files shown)


mlouielu:
It should be used in any modern VTE version (it doesn't invoke any VTE methods), and anyone should able to use it without problem.

gsemet:
Oh ok you don’t use any vte function. I guess this may take some cpu to run, can check if when disabled it really does not have any impact on perf

mlouielu:
It now should cached when loading the same file (e.g. when restoring tabs).

moonlighf:
i love this feature, but your branch lose this feature that search in current terminal

guake/data/org.guake.gschema.xml
```diff
@@ -166,6 +166,16 @@
             <summary>Window vertical alignment.</summary>
             <description>Place Guake at: 0: top, 1: bottom</description>
         </key>
+        <key name="background-image-file" type="s">
+          <default>''</default>
```
la5942: Indents are all off here

guake/utils.py
```diff
@@ -20,12 +20,15 @@
...
 """
 import datetime
+import enum
 import logging
 import os
 import platform
 import subprocess
 import time
 
+import cairo
```
Davidy22, line 30: Does this need to be added to requirements.txt? Looks like a new dependency.

Davidy22 (commented):
Tried it out and was sort of hoping for image transparency, but this is still nice and pretty without that. Would be nice if we used a file picker that showed thumbnails too. Both of those can get added in another commit though.

guake/utils.py
gsemet: requirements.txt is maintained by pipenv

Davidy22: Oh alright, just @la5942's review stands then.

Davidy22:
In addition to polishing off that whitespace objection from @la5942, also going to need to grab the recent CI changes and make sure this passes the build and then we can merge.

Davidy22:
Went and committed the little whitespace edit myself, just some merges and then a passthrough of black left and this should be good to go.

Davidy22 (approved):
Went and did the rebase myself. Still works, passes checks, was already approved before so merging.

PR #1613, opened 2019-06-07
Fix the license classifier in setup.cfg, guake is GPLv2+

pypingou:
Signed-off-by: Pierre-Yves Chibon <pingou@pingoured.fr>

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -10,7 +10,7 @@ home-page = https://github.com/Guake/guake
 classifier =
     Intended Audience :: Information Technology
     Intended Audience :: System Administrators
-    License :: OSI Approved :: GNU General Public License v2 (GPLv2)
+    License :: OSI Approved :: GNU General Public License v2 or later (GPLv2+)
     Operating System :: POSIX :: Linux
     Programming Language :: Python :: 3.5
     Programming Language :: Python :: 3.6


gsemet:
Hi. Just to understand, what is the difference between GPLv2 and v2+? According to https://github.com/Guake/guake/blob/master/COPYING, Guake is GPLv2, without the "and later", isn't?

pypingou:
The COPYING file is the GPLv2, but guake has always been licensed as: "GPLv2 or at your option any later version". So guake itself is not incompatible with the GPLv3 if we so desire to change its license one day.

gsemet:
I would like to better understand on which basis you stand Guake is GPLv2+ despite the licence file saying is GPLv2. For me the licence file give the real licence, it is the source of truth.
I have no complain so far on the GPLv2 and I think it protect the contributors and contribution enough.

I agree GPLv2+ is compatible with GPLv3. However I am not sure if we can change the licence (from GPLv2+ to GPLv3) without having a written agreement of all contributors who ever modified a line in the code. We do not have a CLA for doing that.
So I am not sure if we can even do it in the case we want.

So in short, for me Guake is GPLv2 (that’s its licence file) unless proven otherwise. GPLv2 is ok I do not see the reason why we would take the risk of changing it

pypingou:
> So in short, for me Guake is GPLv2 (that’s its licence file) unless proven otherwise. GPLv2 is ok I do not see the reason why we would take the risk of changing it

Ok, I see the missunderstanding, guake is and has always been GPLv2+, since pretty much the origin. This PR intents to fix the category you've added in 089d72aab3c1708ed3476e42fad7659d8b25fdb7 which is incorrect.

- guake.appdata.xml added in 4153ef38f9044cbed6494075fce80acd5809df2b mentions GPLv2+
- The header in guake/about.py stipulates GPLv2+, same for guake_app.py, main.py
- The file data/about.glade in the tag 0.3.0 mentions GPLv2+

So the category line was wrong, this PR is fixing it back to what guake's license has always been :)

Finally, the COPYING is the text of the GPLv2, there is no text or GPLv2+, you ship the v2 and mention that the project is GPLv2 or, at your option, any later version.

pypingou:
Coming back to this discussion a bit:

> I am not sure if we can change the licence (from GPLv2+ to GPLv3) without having a written agreement of all contributors who ever modified a line in the code. 

That is correct, we cannot change the license of guake itself without the agreement of everyone involved, but piece of code from guake can be incorporated into a GPLv3 project without violating the licensing. That's the interest of the GPLv2+.

gsemet:
ok then, let's merge

pypingou:
Thanks for the review :)

PR #1628, opened 2019-07-29
Add shortcut button to toggle "hide on lose focus" option

pacdouglas:
Hi!

I think this can be useful. Sometimes I need toggle this option and open the "Guake Preferences" all the times... u know? lol. 

Greetings and thanks for your app!

### DIFF
--- a/guake/notebook.py
+++ b/guake/notebook.py
@@ -88,6 +88,9 @@ class TerminalNotebook(Gtk.Notebook):
         )
         self.new_page_button.connect("clicked", self.on_new_tab)
 
+        self.hide_lose_focus_button = Gtk.Button(visible=True)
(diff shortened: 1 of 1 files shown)


gsemet:
hum, I am not fan of clutering the main ui...

pacdouglas:
Sample: 
image

la5942:
> hum, I am not fan of clutering the main ui...

I agree, also not something that all users would want. @pacdouglas Why not use the keyboard shortcut instead of adding unnecessary clutter? `Ctrl+F1`

pacdouglas:
Yeah, I didn't know the shortcut, and on master branch we have a new button to navigate between the tabs, so I was inspired by this buttons. 

But you're right, less is more, I'll close the PR.

PR #1630, opened 2019-08-10
Prefs: Add an option to use the last dir on the path as the tab name.

gordeevbr:
**Context**

This pull request introduces a feature that I wanted for quite some time to improve my workflow.

When I work with Guake, I open up multiple a tab for each project. It looks like this:

Tabs_Before

A bit cramped and not very aesthetically pleasing. Scrolling is inconvenient too, especially since I might have up to 20 of these open.

**What this pull request introduces**

Preferences window now has an option to only display the last directory as a tab name (everything after the last '/' symbol, or (root) if the string is empty). I also merged this option with the abbreviate option into a drop-down menu instead of adding an another checkbox. Like this:

Prefs

With this option on, the tabs look like this:

Tabs_After

Which might be less informative, but far more neat and tidy, in my opinion.

**Considerations**

- The translations should be redone for the preferences menu now.
- I just noticed the tab names aren't updated instantly after you change this option in the preferences menu and then go back to Guake, but rather after Guake is restarted. Was it like that before? I might take a look at this too.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -19,6 +19,8 @@ Bug Fixes
 
 - Fix fullscreen/unfullscreen not handle correctly when trigger by wm
 
+- Fix search revealer causing terminal unclickable at bottom right
+
 - Remove no need window draw callback
 
 - Update some dependencies for build environment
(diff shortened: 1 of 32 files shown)


gsemet:
agreed !

PR #1650, opened 2019-09-27
Add: unfullscreen through D-Bus interface, as well as through cli.

SleepyBag:


gsemet:
Hi. Can you add the reno file (make reno)?

SleepyBag:
Thank you for your reply. Reno file is added now.

PR #1654, opened 2019-10-04
Revamp `make uninstall`

### LINKED ISSUE #1646: "make uninstall" doesn't clean up
`sudo make uninstall` reports the following:

```
rm -f "/usr/local/share/applications/guake.desktop"
rm -f "/usr/local/share/applications/guake-prefs.desktop"
rm -f "/usr/local/share/metainfo/guake.appdata.xml"
rm -fr "/usr/local/share/guake/pixmaps"
rm -fr "/usr/local/share/guake"
rm -f "/usr/local/share/glib-2.0/schemas/org.guake.gschema.xml"
```

First, this list is incompete, it at least removes `/usr/local/bin/guake` which it doesn't report.

Second, more importantly, it leaves tons of installed stuff under `/usr/local`, including `/usr/local/bin/guake-toggle`, `/usr/local/lib/python3.7/dist-packages/guake-3.6.4.dev19-py3.7.egg-info`, `/usr/local/lib/python3.7/dist-packages/guake`, `/usr/local/share/locale/*/LC_MESSAGES/guake.mo`.

As a result, after a manual installation and uninstallation from git, followed by a reinstall of the package provided my distribution (Ubuntu 19.10 beta) I still don't get a usable Guake due to a mixture of two installations (the distro-shipped version tries to load some of the files from `/usr/local`).

Expected behavior: `sudo make uninstall` should remove everything that `sudo make install` installed. Moreover, report everything or nothing (preferably everything) during the process.

ulidtko:
This fixes #1646.

The commit message is repeated below.

-----

`$(prefix)` uppercased for convention: `DESTDIR` and `PREFIX` are the usual parameters of `make install`.
Mixed var-casing just looks clumsy in `make DESTDIR=/tmp/mypkgroot prefix=/usr install`.

Dropped `scripts/find-first-site-package.py` as junk code. It didn't work
altogether, always returning '/'. Because `distutils.sysconfig.get_python_lib()`
returns *one string*, not a list of strings; thus `get_python_lib()[0] == '/'`.

However, it isn't that trivial to detect the correct path to site-packages of
the last install. What we do, is take *all* possible site-packages locations
from `site.getsitepackages([$PREFIX])`, and remove {guake,guake-*.egg-info} from
there. This is robust, and can't do any harm besides wiping [possibly several]
guake packages under the specified PREFIX (which defaults to /usr/local) --
which is exactly what `make uninstall` is supposed to do anyway.

Targets uninstall-old-locale, uninstall-old-schemas dropped. I fail to see any
purpose and/or utility of them. With `OLD_PREFIX:=$(DESTDIR)usr` (which by itself
conflates considerations and simply is wrong), what those targets did can be
trivially emulated via `make DESTDIR=... PREFIX=/usr uninstall`; so there
remains no point in duplicating the uninstall path lists.

Target uninstall-schemas expanded to cover all installed artifacts.

Target uninstall-locale's idempotence fixed using the `|| :` error-silencer.

"Installing from source" warning corrected and made more prominent.

DEV_*_DIR vars dropped as unused, except DEV_DATA_DIR which is used.

Tested with:

    make DESTDIR=/tmp/tt PREFIX=/usr/local install
    make DESTDIR=/tmp/tt PREFIX=/usr/local uninstall
    tree /tmp/tt
    # 13 directories, 0 files


gsemet:
thanks

PR #1675, opened 2019-11-03
Manage colors

Polar-Fox:
Hello there.

It can be useful some times to be able to change colors of the current terminal only (not only of the current tab of whole terminal app).
And I think it's good to have an option to reset colors to the current settings after changing them.

So I added several command line options for managing terminal colors:

**--reset-colors**: Lets the user to reset terminal colors after changing them via **--bgcolor** and **--fgcolor** arguments. Reverts colors to the settings colors.

**--bgcolor-focused**, **--fgcolor-focused**, **--reset-colors-focused**: Changes the background color, the foreground color and resets colors for the currently focused terminal.

### DIFF
--- a/guake/dbusiface.py
+++ b/guake/dbusiface.py
@@ -99,6 +99,10 @@ class DbusManager(dbus.service.Object):
     def change_palette_name(self, palette_name):
         self.guake.change_palette_name(palette_name)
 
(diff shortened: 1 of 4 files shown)


gsemet:
thanks ! look great :)

Polar-Fox:
Thanks for the review!

PR #1676, opened 2019-11-05
Manage colors

Polar-Fox:
Hi!

I renamed **--bgcolor-focused**, **--fgcolor-focused** and **--reset-colors-focused** command line arguments to **--bgcolor-current**, **--fgcolor-current** and **--reset-colors-current** respectively.
Since these commands were added just 2 days ago (https://github.com/Guake/guake/pull/1675), so it seems to be safe to rename them.

Also made above commands to change colors of the **current** terminal, **not** the **focused** one, as they are not the same sometimes (for example in case when these commands are executed within a shell script, and the whole terminal is not focused at the moment).


gsemet:
ok for me.

Polar-Fox:
Thanks!

Issue #1680, opened 2019-11-12
Split Terminals - allow to select terminal like --select-tab in CLI


xurei:
This is related to issue #1645

`--split-vertical` and `--split-horizontal` are really nice ! But it feels to me that a select terminal feature is missing.

Typically, if you want to make a 2x2 grid, you can't right now because the first terminal cannot be selected. Hence you can only split it vertically or horizontally, but not both.

Proposal:
```bash
 guake --select-tab N --select-terminal M
```
Where M would be an auto-incremented value, very much like the tabs work.

I don't know how complicated it would be to implement, but it would make complex startup-scripts more powerful.

Example for a 2x2 grid :

```bash
# Main terminal has index 0
guake --split-horizontally # Right terminal has index 1
guake --split-vertically # Bottom-Right terminal has index 2, Top-Right keeps index 1
guake --select-terminal 0 # Back to left terminal
guake --split-vertically # Bottom-Left terminal has index 3, Top-Left keeps index 0
```

We end up with this layout:
```
|---|---|
| 0 | 1 |
|---|---|
| 3 | 2 |
|---|---|
```

VesterDe:
Thanks for making this!

I'm wondering now, how it would be possible to make a setup script for this 2x2 grid situation?

Let's say I want to run 4 separate commands in those 4 boxes, like vagrant up in one, ssh in another, tail -f in a third etc...

Ideally, I'd write a .sh file for that, but that .sh file will of course execute all commands in the same terminal, regardless of guake focus.

VesterDe:
Ah, I quickly answered my own question.

The -e param works for this.

```
guake --split-horizontal -e "vagrant up --provision && vagrant ssh"
guake --select-terminal 0
guake --split-vertical -e "cd frontend && npm run serve"
```

If anyone needs an example.

xurei:
I believe this issue can be closed now.

PR #1681, opened 2019-11-12
Add --select-terminal and --selected-terminal options in CLI

xurei:
Resolves issue #1680

`--split-vertical` and `--split-horizontal` are really nice ! But it feels to me that a select terminal feature is missing.

Typically, if you want to make a 2x2 grid, you currently can't because the first terminal cannot be selected. Hence you can only split it vertically or horizontally, but not both.

Proposal:
```bash
 guake --select-terminal INDEX
```
Where INDEX would be an auto-incremented value, very much like the tabs work.

Example for a 2x2 grid :
```bash
# Main terminal has index 0
guake --split-horizontally # Right terminal has index 1
guake --split-vertically # Bottom-Right terminal has index 2, Top-Right keeps index 1
guake --select-terminal 0 # Back to left terminal
guake --split-vertically # Bottom-Left terminal has index 3, Top-Left keeps index 0
```

We end up with this layout:
```
|---|---|
| 0 | 1 |
|---|---|
| 3 | 2 |
|---|---|
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,8 @@
 # Linux, IDE and Editors
 
-#    pycharm
+#    pycharm & IntelliJ
 .idea
+*.iml
 
 #    kde
 .directory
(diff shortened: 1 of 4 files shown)


gsemet:
thanks a lot

PR #1704, opened 2020-01-22
fix: hidden scrollbar reappears after new tab

RuRo:
This fixes a small bug, where creating a new tab would temporarily show the scroll bars on other tabs, even when the "Show scrollbar" option is unchecked.


gsemet:
Thanks !

PR #1709, opened 2020-02-08
WIP - Refurbishing code base

pykong:
This **PR** could serve as a starting point for cleaning up and modernizing `Guake`'s code base.
As of 466a9d4 it should be considered work in progress.

Changes included:
- Migrated from  `pipenv` to `poetry`
- Configured `pre-commit`
- Modernized syntax to python 3.6
   -  among other fixes moved all string formatting to *f-string* syntax
- Migrated from `yapf` to `black` and formatted all source files
- Removed and sorted imports via isort
- Various manual fixes and improvements (*see commit history*)


la5942:
Testing and reviewing such a broad range of changes is going to be challenging (this PR is huge). Why not break it down into smaller PR's (as per your PR comment bullets) which are more likely to get review and easier to merge with confidence?

guake/gsettings.py
```diff
@@ -320,7 +320,7 @@ def fstyle_changed(self, settings, key, user_data):
 
         if terminal_uuid:
             terminal = self.guake.notebook_manager.get_terminal_by_uuid(terminal_uuid)
-            terminals = (terminal,) if terminal else tuple()
+            terminals = (terminal,) if terminal else ()
```
gsemet: i think this this quite wrong actually... either we use a list or a `(None,)` but having `terminals` sometimes 1-tuple, sometimes 0-tuple (why does this concept even exist?? it makes no sense).

so I vote to turn `terminals` into list.

gsemet:
Hi. Yes, I recommend to cut it in several part. I am still using on an everyday basis pipenv, despit the lack of support and update. Does poetry good enough to switch ? Does it has a better dependency resolution support (the thing pipenv does on top of pip but bugged as hell) or does it let it completely pip deals with it (so no advanced dependency management)?

I am mostly ok with the current changes, i didn't know flint or pyupgrade. As long as they do not mess with translated string (which would be really hard to migrate to f-string-like syntax) and keep log.debug/log.info with lazy parameters.

Do you want to replace yapf by black? I am ok with that, guake is no more intended into working for python < 3.6

pykong:
@la5942 @gsemet

Thanks for your responses.

Indeed I wanted to give a full picture of changes I like to see, hence this **PR** could be understood as an expose than an actual merge candidate. I wanted to test for your overall support first, as I was suspecting this project no longer under good maintenance (the recent release proving me wrong).
As you appear generally positive let me break this big **PR** into smaller ones.

pykong:
Closing this **PR**, as it is decided to break it down into smaller more manageable ones.
Discussions may continue on the accompanying thread: #1710

Issue #1698, opened 2019-12-22
Refocus not working when calling guake-toggle


pykong:
**Describe the bug**

The refocus function is not working.

**Expected behavior**

With guake's refocus option enabled:
When guake's window is visible and not focused, running `guake -t` or `guake-toggle` should return focus to the open window.

**Actual behavior**
Under said circumstances guake's window gets hidden.

What you actually get.

Described already

**My debugging so far**

I have debugged the reason for guake not refocusing down to this line:

https://github.com/Guake/guake/blob/9d3ef38b6c0fbca3c846e7b55f1a68fccdbd6f8f/guake/guake_app.py#L570

where ~~`self.losefocus_time` and~~ `event_time` will always be **0** no matter what.

Setting `event_time` to an arbitrarily high number so the statement evaluates true if `self.losefocus_time > 0` keeps the window open at all times and it can neither be closed nor focused via the normal toggle command (albeit by other means such as mouse).

**Other issues**

#1651
#1188
#1212
-------------------------------------------------------------------------------

<details><summary>$ guake --support</summary>

Guake Version:		3.6.3

Vte Version:		0.52.2

Vte Runtime Version:	0.52.2

--------------------------------------------------
GTK+ Version:		3.22.30

GDK Backend:		GdkX11.X11Display

--------------------------------------------------
Desktop Session: cinnamon

--------------------------------------------------
Display: :0

RGBA visual: True

Composited: True

* Monitor: 0 -  eDP-1-1
    * Geometry:		1920 x 1080 at 932, 1600
    * Size:		344 x 193 mm²
    * Primary:		True
    * Refresh rate:	60.01 Hz
    * Subpixel layout:	unknown
* Monitor: 1 -  DP-1-3
    * Geometry:		3840 x 1600 at 0, 0
    * Size:		880 x 367 mm²
    * Primary:		False
    * Refresh rate:	59.99 Hz
    * Subpixel layout:	unknown

pykong:
@gsemet If you gave me some useful input on the issue I might be able to bring forth a PR.

pykong:
**Further analysis:**
I can now confirm that the refocus function indeed works!
However, it has to be toggled via the shortcut setup in `guake` itself, i.e. F12.

In contrast I have been toggling `guake` by a system shortcut calling `guake-toggle`.

pykong:
Modifying the `show_hide` method in the following way perfectly conducts `refocus` when toggling via `guake-toggle`.

@gsemet 
[...]

Issue #1710, opened 2020-02-08
Modernize and Overhaul Code Base


pykong:
It makes me sad to see this great project slowly dying due to a lack of maintenance.

As of writing this, there are no less than 390 open issues on this repository.
Its [Feature Request](https://github.com/Guake/guake#features-request) displays 72 open requests with the most painfully needed request of an official Launchpad PPA counting 181 votes and not being addressed for a whole year now. The repository itself has seen the last commit in Mid December 2019, with 24 stale branches rotting away undeleted and uncared for.

To get out of this rut significant dedication would be required in the form of core and new contributors. To attract fresh contributors, I suggest modernizing and cleaning up the codebase to make it more inviting.

My **PR** #1709 might serve as a starting point in this regard. It includes the following adaptions:

- Moving from `pipenv` to `poetry`, as developing with the latter is so much more pleasant, just talking dependency resolution here. This includes the addition of a `pyproject.toml` file.
- Moving from `yapf` to `black` for formatting, as it is what everyone seems to be using nowadays and I consider a significantly better tool.
- Sort and cleanup imports.
- Dropping Python 3.5 support as it reaches EOL this year anyway, allowing us to use features of Python 3.6 syntax, especially f-strings. (Syntax has been modernized  via [pyupgrade](https://github.com/asottile/pyupgrade))
- Using [pre-commit](https://github.com/pre-commit/pre-commit) for managing git hooks.

Some further suggestions not covered in that **PR** are:
- Move from Travis CI to GitHub Actions for testing and deployment (new contributors would not need to open Travis CI account for running tests against their fork.)
- Major refactoring. We should strive to break down the monolithic structure into smaller more understandable pieces.
- ~~Using PyTest for all things testing, as it is simply the best test suite for python.~~ **EDIT:** It already does.
- Type annotate entire code base, which gives us type safety and also might pave a way for increasing performance via [mypyc](https://github.com/python/mypy/tree/master/mypyc)

gsemet:
hi ! thanks for your dedication. I mostly agree on what you are saying, so you can count on me to help you integrating your changes. I have less and less time to maintain guake myself, and indeed there are some rotting stuff arround that need clean up. First I can close old pull requests where i have no more anwer on it, and maybe delete older branches.

[...]

PR #1711, opened 2020-02-08
Re-implement Refocus

pykong:
This **PR** modifies the `Guake.show_hide` method in order to re-enable the refocus function.
It works perfectly well on my Linux Mint 19.3 with Cinnamon. 

In #1698 @gsemet suggested to test a large range of different desktop environments with this modification.
In absence of automatic stets alleviating the extraordinary manual effort required to do so, I am instead suggestion an alternative approach:
**Marking this feature as experimental / test via pre-release.**
Any bug issues could then be solved as they occur.

Also the code base could use some overhaul, as I have suggested in #1710 - regarding the **PR** at hand I have not dared to examine a possible unification of the different focus methods such as:
- `self.window.get_window().focus(0)`
- `self.set_terminal_focus()`
- `show_focus`

I have not followed the usual steps described for making **PR**s, such as generating a slug file, as I find them difficult to do.

Closes: #1698

### DIFF
--- a/releasenotes/notes/Reimplemented-Refocus-6fdd3242caf8b388.yaml
+++ b/releasenotes/notes/Reimplemented-Refocus-6fdd3242caf8b388.yaml
@@ -1,7 +1,0 @@
-release_summary: >
-  Re-implemented the refocus functionality.
-
-  This functionality allows the user to return the focus to an open guake window.
-  It happened to be partially be maintained but faced issues in the migration to Gtk3.
-
-  The functionality was revived, partially based on commented-out code found in Guake.show_hide().
--- a/releasenotes/notes/Reimplemented-Refocus-6fdd3242caf8b388.yaml
+++ b/releasenotes/notes/Reimplemented-Refocus-6fdd3242caf8b388.yaml
@@ -1,0 +1,7 @@
+release_summary: >
+  Re-implemented the refocus functionality.
+
+  This functionality allows the user to return the focus to an open guake window.
+  It happened to be partially be maintained but faced issues in the migration to Gtk3.
+
+  The functionality was revived, partially based on commented-out code found in Guake.show_hide().
--- a/guake/guake_app.py
+++ b/guake/guake_app.py
@@ -540,6 +540,7 @@ class Guake(SimpleGladeApp):
         """Toggles the main window visibility
         """
         log.debug("Show_hide called")
+        
         if self.forceHide:
             self.forceHide = False
             return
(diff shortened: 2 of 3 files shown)


gsemet:
Hi. I recommend to add a switch in the preference panel to enable the "experimental algorithm", so that users can switch to them on their own. On a futur version we will make it the default while keeping old algo available in case of.

as for the slug, it makes release easier. `make reno SLUG=experimental` and you add a one-line to it. I also tried towncrier which is on the same kind of release note helper, but i have added lot of scripts on top of reno, to generate the release note, changelog etc. But feel free to propose a better alternative :)

pykong:
@gsemet OK. I am implementing the slug,

Regarding the switch: Do you have/want a switch activating all possible experimental features, or one specifically for the refocus function?
Most importantly the refoxus function itself is already activatable by a switch in the preferences menu - you want me to add another one?

Other question: Is there any way you would suggest we write an automatic test for the behaviour?

gsemet:
I must admit, I do a lot of pytest on a daily basis, but testing guake was always really hard, because lot of thing could break from one system to another... so unit tests are pretty useless... except to protect the code on refacto.

But, if we protect ourself with a major release of Guake, we can leave user on old systems with older version of guake and reimplement a complete algorithm and do our best to support major systems such as the latest ubuntu, fedora. The transition to gtk3 has been made like this, and nobody complained.

I guess for the focus/refocus algorithm we can ditch the old code and reimplement it without feature switch. So simply add the slug and I will test it on ubuntu later.

Maybe for more complex algo like window positionning or stuff like this will require feature switch.

pykong:
@gsemet I have now added a slug file.
Please check. Is there anything more to do for me regarding this **PR**.

releasenotes/notes/Reimplemented-Refocus-6fdd3242caf8b388.yaml
```diff
@@ -0,0 +1,7 @@
+release_summary: >
```
gsemet: please use the `features` key:

```yaml
features:
  - |
    Re-implemented the refocus functionality.

    This functionality allows the user to return the focus to an open guake window.
     It happened to be partially be maintained but faced issues in the migration to Gtk3.
    The functionality was revived, partially based on commented-out code found in Guake.show_hide().
```
The > syntax in yaml will gather all your multiple lines to a single one. I use the `|` syntax.

pykong:
@gsemet Please have a look. Is there anything left for me to do before you will merge this **PR**?

gsemet:
look fine :) let's try it now !

PR #1724, opened 2020-02-16
WIP - Migrate to poetry

pykong:
This **PR** will modernize Guake's package management by switching from `pipenv` to `poetry`, as discussed under #1710.

## TODO:
- [ ] make Travis run green
- [ ] put some content into slug-file

### Check:
- [ ] are the updates to the docs in the right place _(gtk3.rst)_?
- [ ] checked the complete build process _(beyond my means)_
- [ ] check the changed generation of the requirements files. Is it OK that:
    - no preamble in files any longer _(No "do not modify")_
    - `requirements-dev.txt` now contains all dependencies, including prod ones

## Done:
- introduced and populated `pyprpject.toml`
- replaced all _pipenv_ commands with their respective _poetry_ counterparts, _(e.g. in MAKEFILE)_
- eliminated all occurrences of `pipenv` except within release notes
- upped version to `4.0.0` _(it is my understanding we aim for a bigger release)_
- ` make style check reno-lint test` does fail because of error within *flake8* itself

## Links:
- https://python-poetry.org/docs/


gsemet:
you have to test the build and the versionning with pbr, the most important stuff is to keep the use of setuptools or to use https://github.com/mtkennerly/poetry-dynamic-versioning. 
pbr does some things:
- version comes from the git history (previous tag). When built, the version is kept in a metadata in the package
- inject de dependencies from requirements.txt to setup.py

pyproject.toml
```diff
@@ -0,0 +1,48 @@
+[tool]
+[tool.poetry]
+name = "guake"
+version = "4.0.0"
```
gsemet: try to use https://github.com/mtkennerly/poetry-dynamic-versioning

pyproject.toml
```diff
@@ -0,0 +1,48 @@
+[tool]
```
gsemet: no build-system part?

requirements.txt
```diff
@@ -1,14 +1,2 @@
...
-# This has been done to maintain backward compatibility with tools and services
-# that do not support `Pipfile` yet.
-#
-# Do NOT edit it directly, use `pipenv install [-d]` to modify `Pipfile` and
-# `Pipfile.lock` and then regenerate `requirements*.txt`.
-################################################################################
-
-pbr
-typing ; python_version < '3.5'
+pbr==5.4.4
```
gsemet: if we use pbr, it is important to keep requirement.txt not frozen (without the version)

pykong:
@gsemet Thanks for your comments.
Can you have a look into Travis logs?

As far as I understand that is the only part still failing.
As you know your Makefile best - do you got any idea on this?
https://travis-ci.org/Guake/guake/jobs/651197596#L4140-L4176

pyproject.toml
pykong: I have not much experience with build systems.
Do you got any idea what to put there?

pyproject.toml
pykong: I was under the impression we were working up to a major release. See your comment here:
https://github.com/Guake/guake/pull/1711#issuecomment-584732273

But no problem. Should I put the version at `3.7.0`then?

requirements.txt
pykong: I do not think there is a straightforward way to generate the requirements files while leaving the version of any dependency blank.

Is `poetry-dynamic-versioning` the plugin you mentioned? Should we try to replace `pbr` with that plugin?

pyproject.toml
gsemet: That’s why I think deriving version from the got history is much powerful. The day I do the release I can choose whatever the version I think it is best.

So my answer would be: don’t put any version here, check with the poetry plugin if it does the job

pykong:
> * inject de dependencies from requirements.txt to setup.py

Can you have a look into [dephell](https://github.com/dephell/dephell)?
Maybe it would suit this purpose well.

pykong:
@gsemet I am afraid I can not get `poetry-dynamic-versioning` to install and tank the entire subsequent build process. Have you got any suggestions?

pykong:
@gsemet I am sorry but the whole build setup behind Guake is beyond my understanding.
I am afraid I can no longer proceed with this **PR**. Feel free to reopen this PR in order to continue working on it.

I am going to focus on: https://github.com/Guake/guake/pull/1711 instead.

Issue #1350, opened 2018-06-28
Two monitors bad terminal width alignment


damianw345:
While having dock on the left side Guake is partially hidden below dock.
While having dock on the right side Guake is partially on the left monitor.

System information:
* Ubuntu 18.04
* Gnome version: 3.28.1

$ guake --version
Guake Terminal: 3.3.1
VTE: 0.52.2
VTE runtime: 0.52.2
Gtk: 3.22.30

![dockleft](https://user-images.githubusercontent.com/26282773/42061837-c5dd427a-7b2b-11e8-9fcb-f866621916f7.jpg)
![dockright](https://user-images.githubusercontent.com/26282773/42061838-c5f9ad84-7b2b-11e8-81b8-3a859fceaae5.jpg)

aichingm:
Have you tried playing the the displacement options?

Preferences -> Main Window

Reduce the width, set the position to left or right and add some pixels of displacement.

gsemet:
i think the window calculation can be completely redone in a futur version... but not for today

obrejla:
Same problem...it's quite ugly :/

priestoferis:
Yup, ran into this as well :( For me it actually works ok on the larger, non-main screen which is on the left side, and messes it up on the smaller main screen on the right.

frz-dev:
Similar problem here:
Left monitor (with dock): guake goes beyond the screen limit and ends up showing on the right monitor
Right monitor (no dock): guake shows ok

@aichingm The alignment/displacement options seem to be completely ignored

frz-dev:
I managed to solve the issue in my case.
The problem was that ubuntu now uses its own unity-like dock, which is an extension of gnome-shell.
As such, the dock is not correctly detected nor managed by guake, as it only works with the unity dock.

I wrote the code to handle the gnome dock:
In `guake_app.py`:
Before `set_final_window_rect`:
```
    def is_using_gnome(self):
        if os.environ.get('XDG_CURRENT_DESKTOP', '').lower() == "gnome".lower():
            return True

        if os.environ.get('XDG_CURRENT_DESKTOP', '').lower() == "ubuntu:gnome".lower():
            return True

        return False
```

In `set_final_window_rect`, after the `if self.is_using_unity()` block:
```
        if self.is_using_gnome():
            #log.debug("CURRENT MONITOR: %s", monitor)
            try:
                dock_enabled = subprocess.check_output([
[...]

PR #1729, opened 2020-02-24
Manage window resizing with Gnome Shell dock (Issue #1350)

frz-dev:
Issue #1350 is due to the fact that Gnome shell dock, currently used by Ubuntu, is not managed.
This causes the window to be badly sized and/or positioned.

This pull request correctly resize/position the Guake window in the presence of a Gnome shell dock.

### DIFF
--- a/guake/guake_app.py
+++ b/guake/guake_app.py
@@ -1113,83 +1113,96 @@ class Guake(SimpleGladeApp):
 
         if self.is_using_gnome():
             #log.debug("CURRENT MONITOR: %s", monitor)
+            dock_enabled = False
             try:
                 dock_enabled = subprocess.check_output([
-                        '/usr/bin/dconf', 'list', '/org/gnome/shell/extensions/dash-to-dock/'
+                    '/usr/bin/dconf', 'list', '/org/gnome/shell/extensions/dash-to-dock/'
+                ])
+            except Exception as e:
+                pass
+
+            ## If dock is enabled, check if it's active on current monitor ##
+            dock_active = False
+            if dock_enabled:
+                log.debug("Gnome dock enabled")
+        
+                #dconf VALUES:
+                #preferred-monitor=0|1 : monitor_num on which dock is shown
+                #multi-monitor=true|false : true=>shown on all monitors
+                #dock-fixed=true|false : false=>auto-hide on , unset=>true
+                #dock-position='LEFT'|'RIGHT'|'BOTTOM' (unset = 'LEFT')
+                #dash-max-icon-size=n : size in pixels
+
+                #check if active on all monitors
+                try:
+                    multi_monitor = subprocess.check_output([
+                        '/usr/bin/dconf', 'read',
+                        '/org/gnome/shell/extensions/dash-to-dock/multi-monitor'
+                    ]).decode("utf-8").rstrip() or "true"
+                except Exception as e:
+                    pass
+
+                if multi_monitor == "true":
+                    dock_active = True
+                else:
+                    try:
+                        preferred_monitor = subprocess.check_output([
+                            '/usr/bin/dconf', 'read',
+                            '/org/gnome/shell/extensions/dash-to-dock/preferred-monitor'
+                        ])
+                        #check if we are on preferred monitor
+                        if monitor == int(preferred_monitor):
+                            dock_active = True
+                    except Exception as e:
+                        pass
+
+            ## If active, check if and where the dock is shown ##
+            dock_shown = False
+            if dock_active:
+                log.debug("Gnome dock active")
+                #check if auto-hide is on
+                dock_fixed = "true"
+                try:
+                    dock_fixed = subprocess.check_output([
+                        '/usr/bin/dconf', 'read',
+                        '/org/gnome/shell/extensions/dash-to-dock/dock-fixed'
+                    ]).decode("utf-8").rstrip() or "true"
+                except Exception as e:
+                    pass
+
+                #if dock-fixed is unset is considered true
+                if dock_fixed != "false":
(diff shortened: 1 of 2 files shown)


guake/guake_app.py
```diff
@@ -1004,6 +1004,17 @@ def is_using_unity(self):
                 return True
         return False
 
+    def is_using_gnome(self):
+        linux_distrib = platform.linux_distribution()
+        if float(linux_distrib[1]) - 0.01 < 11.10:
+            if os.environ.get('XDG_CURRENT_DESKTOP', '').lower() == "gnome".lower():
+                return True
+        else:
+            if os.environ.get('XDG_CURRENT_DESKTOP', '').lower() == "ubuntu:gnome".lower():
```
ulidtko, line 1013: `"ubuntu:gnome".lower()` is the same as just `"ubuntu:gnome"`; likewise for `"gnome"`

guake/guake_app.py
```diff
@@ -1096,6 +1107,88 @@ def set_final_window_rect(self):
...
+                                        '/org/gnome/shell/extensions/dash-to-dock/dash-max-icon-size'
+                                    ])
+                                )
+                                ## Adjust width ##
+                                log.debug("Adjusting window size")
+                                dock_shown = True
+                                dock_size += 19
+                                window_rect.width = window_rect.width - dock_size
+                                if monitor == 1:
+                                    window_rect.x += dock_size
```
ulidtko, line 1186: This is what, **nineteen** nesting levels?..

That's too bad; please see if you can refactor the code into something neater.

guake/guake_app.py
ulidtko: I understand that you copied the version comparison logic from Gaetan's Unity-related code. Nevertheless, **casting version strings to float** isn't going to fly very far; that's simply incorrect. With versions, it isn't true that `11.2 > 11.10`. Either add a big fat `FIXME` right here (and there with Unity), or take e.g. [semver lib](https://pypi.org/project/semver/) and do it right.

ulidtko: Next piece of copy-paste. Could you make a variable here?
```python
def is_using_gnome(self):
    linux_distrib = platform.linux_distribution() [1]
    xdg_desktop = os.environ.get('XDG_CURRENT_DESKTOP', '').lower()
    return (distro_before_11_10 and xdg_desktop == "gnome"
          or distro_after_11_10 and xdg_desktop == "ubuntu:gnome")
```

frz-dev: As for version comparison, casting to float is not a problem for Ubuntu as its version number always have a trailing zero in case of one-digit number (so it would be 11.02, not 11.2.)

Anyway, semver has problems with trailing zeroes (it raises a ValueError exception).
I found this [packaging.version](https://packaging.pypa.io/en/latest/version/#packaging.version.parse) module that works just fine (11.04 < 11.10 yields True, 11.2 > 11.10 yields False)

I assume `packaging` should be added to the requirements, but I'm not sure how this is handled.
Should I add this line:  
```packaging = "*"```
to `Pipfile` ?

Comment (frz-dev) [reply guake/guake_app.py]: As for the variable, I think I made a mistake in using `XDG_CURRENT_DESKTOP` in both cases.
It should be `DESKTOP_SESSION` if version is lower than 11.10.

Also, I think that `linux_distrib[1]) - 0.01` was actually meant to actually include 11.10, so I'm changing `<` for `<=`.

Comment (frz-dev) [reply guake/guake_app.py]: I pushed changes to `is_using_gnome`. I'll wait for your feedback to resolve the conversation

File: guake/guake_app.py (hunk 2 above)
Comment (frz-dev) [reply guake/guake_app.py:1190]: Ok, I'll check it tomorrow

Comment (frz-dev) [reply guake/guake_app.py]: I commited the new version with less nesting levels. 
Basically I serialized all the if statements. Don't really like it conceptually, but I guess it's a matter of taste.

Please note that the new code is UNTESTED
(I'm confined at home for the coronavirus, so I'm working on different machine with no double monitor).

File: guake/guake_app.py [hunk 3]
```diff
@@ -1004,6 +1005,20 @@ def is_using_unity(self):
                 return True
         return False
 
+    def is_using_gnome(self):
+        linux_distrib = platform.linux_distribution()
+        if linux_distrib[0].lower() != "ubuntu":
```
Davidy22, line 1010: Is this function's logic correct? I'm sure there's distros other than ubuntu out there that use GNOME, and I think ubuntu switched back to gnome shell at some point later. Is there no better way to check for gnome shell?

Davidy22 (changes requested):
I don't know enough about gnome shell to comment on the second half with implementation specifics and I can't test this either because I don't use it. Going to need someone else to test and confirm that this code works before merging.

The CI pipeline has been fixed so you'll need to update from master to have our checks run on your branch, and you'll need to get your code to pass our automated tests as well.

Also, would be nice if you could fill in a release notes file describing your fixes, type:

```
make reno SLUG=fix_gnome_shell_window_resize
```

in your shell to open a template in your text editor, delete the sections that aren't relevant to your changes.

gsemet:
can we merge this ticket?

Davidy22:
There's some strange checking logic for detecting gnome shell that I wanted clarification on. I haven't tested this but [this guy did](https://github.com/Guake/guake/pull/1765#issuecomment-920054239) and says this doesn't fix the issue. It's also slightly worrying that the submitter's last messages are that this code is untested, and it stops there.

Davidy22:
We've had a confirmation that this pull request doesn't work, there's still some quality issues and no replies for a while. Closing pr, if you have an updated version of this pull request that does work in the future we can take another look at this.

Issue #1703, opened 2020-01-20
Please re-add original author to docs and license


gabrielfalcao:
Hello Guake maintainers,

I'd like to start by thanking you for continuously giving love to Guake and keeping it an amazing tool for lovers of GTK.

I'm the original author of the project, it started in 2007 as the [guake-gnome-vte sourceforge project](https://sourceforge.net/projects/guake-gnome-vte/) (before github was mainstream)

Around 2009 I stopped using Linux for desktop  and contributing to Guake. 
https://github.com/Guake/guake/commits?author=gabrielfalcao.

At some point my name was removed from the headers of *most* python files as well as from the list of authors.

You can see the original and old commits here: https://github.com/gabrielfalcao/guake/blob/master/AUTHORS

The original code (still distributed) with my name on the "about window": https://github.com/Guake/guake/blob/d75d29022a890c26ce806107f9789faee83ef469/guake/about.py#L4

This is very unfortunate because Guake was one of my greatest creations and became a living organism, but without proper credit it becomes hard to prove that I've built something valuable to the world.

## What I ask

Please kindly add a small footnote to README.rst and to the docs website so that I can get proper credit.

I intend to apply for a visa and would benefit from having my name mentioned somewhere else than the "About" window of the application, that way lawyers can refer to that link rather than having to install the application.

Thank you

oelmekki:
FYI, this was posted [on hacker news](https://news.ycombinator.com/item?id=22523814).

AmazingTurtle:
The file was deleted in 2017 in [this](https://github.com/Guake/guake/commit/089d72aab3c1708ed3476e42fad7659d8b25fdb7#diff-3d350169560e75d0cf9fc8e3574a3639) commit by @gsemet

gsemet:
Hi all and hi @gabrielfalcao ! AUTHORS file is actually automatically generated by PBR using the git history, so all authors are automatically synced when building the wheel. So the file does not need to be committed.
This list is as of today the following:
```
A S Alam <apreet.alam@gmail.com>
Adolfo Correa <fitomillennium@hotmail.com>
Adonis K <aklp08@gmail.com>
Airfunker <airfunker@gmail.com>
Aleksandar Krsteski <alekrsteski@gmail.com>
Aleksandr Mezin <mezin.alexander@gmail.com>
Alessio Garzi <agarzi@clouditalia.com>
Alessio Garzi <gun101@email.it>
Alex Shkop <ashkop@sciforce.solutions>
[...]

PR #1733, opened 2020-03-09
Mention original author

tuxillo:
Fixes issue #1703


tuxillo:
I'm looking at the documentation too, not sure where the mention should go. Any ideas?

gabrielfalcao:
> I'm looking at the documentation too, not sure where the mention should go. Any ideas?

I'd be glad to write the original history here: https://github.com/Guake/guake/blob/master/docs/source/project/history.rst

I added a comment on HW: https://news.ycombinator.com/item?id=22524552

gsemet:
Thanks for this doc update. Do not hesitate to take over the history part of Guake, I do not know it actually !

PR #1786, opened 2020-07-21
Save tabs when directory changed, not `on_terminal_title_changed` (#1633)

### LINKED ISSUE #1633: save_tabs function is way too aggressive
On current master (), the save_tabs function is being called way too much.
When PS1 is set to update the title, save_tabs is called every time Enter is pressed in guake.

To reproduce:
Add this to .bashrc, or just run this snippet in a guake shell.
```
# If this is an xterm set the title
case "$TERM" in
xterm*|rxvt*)
    PS1="\[\e]0;\W\a\]$PS1"
    ;;
*)
    ;;
esac
```
Press enter in guake terminal for about a second:
Guake tabs saved to /home/user/.config/guake/session.json
mom, I've been called: on_terminal_title_changed <function Guake.on_terminal_title_changed at 0x7fd9a0196ea0> at 1566561134
Guake tabs saved to /home/user/.config/guake/session.json
mom, I've been called: on_terminal_title_changed <function Guake.on_terminal_title_changed at 0x7fd9a0196ea0> at 1566561135
Guake tabs saved to /home/user/.config/guake/session.json
mom, I've been called: on_terminal_title_changed <function Guake.on_terminal_title_changed at 0x7fd9a0196ea0> at 1566561135
Guake tabs saved to /home/user/.config/guake/session.json
mom, I've been called: on_terminal_title_changed <function Guake.on_terminal_title_changed at 0x7fd9a0196ea0> at 1566561135
Guake tabs saved to /home/user/.config/guake/session.json
mom, I've been called: on_terminal_title_changed <function Guake.on_terminal_title_changed at 0x7fd9a0196ea0> at 1566561135
Guake tabs saved to /home/user/.config/guake/session.json
mom, I've been called: on_terminal_title_changed <function Guake.on_terminal_title_changed at 0x7fd9a0196ea0> at 1566561135
Guake tabs saved to /home/user/.config/guake/session.json
mom, I've been called: on_terminal_title_changed <function Guake.on_terminal_title_changed at 0x7fd9a0196ea0> at 1566561135
Guake tabs saved to /home/user/.config/guake/session.json
mom, I've been called: on_terminal_title_changed <function Guake.on_terminal_title_changed at 0x7fd9a0196ea0> at 1566561135
Guake tabs saved to /home/user/.config/guake/session.json
mom, I've been called: on_terminal_title_changed <function Guake.on_terminal_title_changed at 0x7fd9a0196ea0> at 1566561135
Guake tabs saved to /home/user/.config/guake/session.json
mom, I've been called: on_terminal_title_changed <function Guake.on_terminal_title_changed at 0x7fd9a0196ea0> at 1566561135
Guake tabs saved to /home/user/.config/guake/session.json
mom, I've been called: on_terminal_title_changed <function Guake.on_terminal_title_changed at 0x7fd9a0196ea0> at 1566561135
Guake tabs saved to /home/user/.config/guake/session.json
mom, I've been called: on_terminal_title_changed <function Guake.on_terminal_title_changed at 0x7fd9a0196ea0> at 1566561135
Guake tabs saved to /home/user/.config/guake/session.json
mom, I've been called: on_terminal_title_changed <function Guake.on_terminal_title_changed at 0x7fd9a0196ea0> at 1566561135
Guake tabs saved to /home/user/.config/guake/session.json
mom, I've been called: on_terminal_title_changed <function Guake.on_terminal_title_changed at 0x7fd9a0196ea0> at 1566561135
Guake tabs saved to /home/user/.config/guake/session.json
mom, I've been called: on_terminal_title_changed <function Guake.on_terminal_title_changed at 0x7fd9a0196ea0> at 1566561135
Guake tabs saved to /home/user/.config/guake/session.json
mom, I've been called: on_terminal_title_changed <function Guake.on_terminal_title_changed at 0x7fd9a0196ea0> at 1566561135
Guake tabs saved to /home/user/.config/guake/session.json
mom, I've been called: on_terminal_title_changed <function Guake.on_terminal_title_changed at 0x7fd9a0196ea0> at 1566561135
Guake tabs saved to /home/user/.config/guake/session.json

mlouielu:
This should fix #1633.

This commit remove `@save_tabs_when_changed` on `on_terminal_title_changed`,
and put the terminal directory into ther instance. This is because
`vte:current-directory-uri-changed` is not worked, and `current-directory-uri`
always return `None`.

When `on_terminal_title_changed` called, it will check if the directory
changed for this terminal, if so, then save the tabs by the setting.

### DIFF
--- a/guake/guake_app.py
+++ b/guake/guake_app.py
@@ -1097,7 +1097,17 @@ class Guake(SimpleGladeApp):
             pass
         return TabNameUtils.shorten(vte_title, self.settings)
 
-    @save_tabs_when_changed
+    def check_if_terminal_directory_changed(self, term):
+        @save_tabs_when_changed
+        def terminal_directory_changed(self):
+            # Yep, just used for save tabs when changed
+            pass
+
+        current_directory = term.get_current_directory()
+        if current_directory != term.directory:
+            term.directory = current_directory
+            terminal_directory_changed(self)
+
     def on_terminal_title_changed(self, vte, term):
         # box must be a page
         if not term.get_parent():
@@ -1124,6 +1134,9 @@ class Guake(SimpleGladeApp):
             text = nb.get_tab_text_page(box)
             if text:
                 self.update_window_title(text)
+
+        # Check if terminal directory has changed
+        self.check_if_terminal_directory_changed(term)
 
     def update_window_title(self, title):
         if self.settings.general.get_boolean("set-window-title") is True:
@@ -1161,6 +1174,9 @@ class Guake(SimpleGladeApp):
         terminal.handler_ids.append(
             terminal.connect("window-title-changed", self.on_terminal_title_changed, terminal)
         )
+
+        # Use to detect if directory has changed
+        terminal.directory = terminal.get_current_directory()
 
     @save_tabs_when_changed
     def add_tab(self, directory=None):
--- a/guake/guake_app.py
+++ b/guake/guake_app.py
@@ -1112,6 +1112,10 @@ class Guake(SimpleGladeApp):
         # box must be a page
         if not term.get_parent():
             return
+
+        # Check if terminal directory has changed
+        self.check_if_terminal_directory_changed(term)
+
         box = term.get_parent().get_root_box()
         use_vte_titles = self.settings.general.get_boolean("use-vte-titles")
         if not use_vte_titles:
@@ -1134,9 +1138,6 @@ class Guake(SimpleGladeApp):
             text = nb.get_tab_text_page(box)
             if text:
                 self.update_window_title(text)
-
-        # Check if terminal directory has changed
-        self.check_if_terminal_directory_changed(term)
 
     def update_window_title(self, title):
         if self.settings.general.get_boolean("set-window-title") is True:
--- a/guake/guake_app.py
+++ b/guake/guake_app.py
@@ -562,7 +562,7 @@ class Guake(SimpleGladeApp):
             self.set_terminal_focus()
             return
 
-        should_refocus = self.settings.general.get_boolean('window-refocus')
+        should_refocus = self.settings.general.get_boolean("window-refocus")
         has_focus = self.window.get_window().get_state() & Gdk.WindowState.FOCUSED
         if should_refocus and not has_focus:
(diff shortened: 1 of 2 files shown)


mlouielu:
This may not work for split window...

mlouielu:
> I'm somewhat dubious of the overall approach here. It seems to rest on the confusion that VTE titles are a representation of the CWD (current working directory). They aren't.
> 

This PR didn't use VTE titles as their CWD, here we only use `on_terminal_title_changed` as a signal that "terminal directory may have changed"

> It just so happens that `bash` on many systems by default sends `[{{cwd}}]` as the VTE title, and many people don't [customize that](https://github.com/ulidtko/versioned-configs/blob/master/HOME/zshrc#L117,L122), neither interact with [other programs](https://askubuntu.com/q/22413/9061) which set VTE titles too. Still many _other_ people do do that, and those will see why this PR misses the point.
> 

I've test with default bash shell, and `Vte.Terminal.get_current_directory()` will changed when the directory changed.

Or are there any cases that `Vte.Terminal.get_current_directory()` will failed even if directory changed? If so, then the `save-tabs` function will not work too. Because the path that save to the `session.json` also use `Vte.Termianl.get_current_directory()`.

> So instead of adding more logic entrenching the confusion — some logic needs to be _removed_. Specifically, as the original issue reporter said:
> 
> > there is no reason to save the "label" unless when "custom_label_set": true. It will anyway not be used when restoring the tabs.
> 
> Please follow this. When a program sends to Guake _please name my tab "IAmFoobar"_ — there's no slightest need to persist `"IAmFoobar"` anywhere; the program will send it again. When _the user_ however names a tab "MyAwesomeSpecialTab" — we want to ignore what the programs send, and we want to save this manual label into `session.json` and restore it when restarted.

Again, this PR only use `on_terminal_title_changed()` signal as a signal that terminal directory may changed.


* `@save_tabs_when_changed` has been removed from `on_terminal_title_changed()`
https://github.com/Guake/guake/pull/1786/files#diff-e41ec8d6a3b1c2088836632a5c13ec40L1100

*  `self.check_if_terminal_directory_changed(term)` will check if the terminal directory changed or not.

Issue #1820, opened 2020-12-21
Inconsistent window sizing and position in dual-monitor setups under Wayland


bitscher:
**Describe the bug**
Inconsistent window sizing between monitors in dual-monitor setup with different resolutions on Gnome 3.38.2 Wayland on Fedora 33.

Modifying the environment variable  before calling guake by prepending `env GDK_BACKEND=x11` fixes the issue which points to an issue with the existing attempt at overriding the value of `GDK_BACKEND` in the code.
https://github.com/Guake/guake/blob/9c213270767687b30108a7206ac4f4bb913385cd/guake/main.py#L57-L67

**Expected behavior**
Window size should be correct on both monitors.

**Actual behavior**
Window size is always the same regardless of the resolution of the monitor it's being displayed on. The position is also offset if monitors are not horizontally aligned from the top.

Monitor 0 (1080p)
![image](https://user-images.githubusercontent.com/4034221/102729782-fed1b680-42e6-11eb-96e2-c45093f859bf.png)

Monitor 1 (1440p)
![image](https://user-images.githubusercontent.com/4034221/102729700-a7cbe180-42e6-11eb-9b3e-33031320d788.png)

**To Reproduce**
- On a dual monitor setup with different monitor resolutions
- Run Gnome-shell on Wayland
- Open the terminal on monitor 1
- Close the terminal, and open it on monitor 2

<details><summary>$ guake --support</summary>

Guake Version:		3.6.3

Vte Version:		0.62.1

Vte Runtime Version:	0.62.1

--------------------------------------------------
GTK+ Version:		3.24.24

GDK Backend:		GdkX11.X11Display

--------------------------------------------------
Desktop Session: gnome

--------------------------------------------------
Display: :0

RGBA visual: True

Composited: True

* Monitor: 0 -  XWAYLAND0
    * Geometry:		1920 x 1080 at 0, 239
    * Size:		530 x 300 mm²
    * Primary:		True
    * Refresh rate:	143.88 Hz
    * Subpixel layout:	unknown
* Monitor: 1 -  XWAYLAND1
    * Geometry:		2560 x 1440 at 1920, 0
    * Size:		600 x 340 mm²
    * Primary:		False
    * Refresh rate:	143.91 Hz
    * Subpixel layout:	unknown

PR #1821, opened 2020-12-21
Fix ineffective override of GDK_BACKEND envvar

bitscher:
https://github.com/Guake/guake/blob/9c213270767687b30108a7206ac4f4bb913385cd/guake/main.py#L57-L67 does attempt to override the `GDK_BACKEND` environement variable at line 67 in main.py but it is already too late since `Gdk` is imported through earlier imports from `guake.support` and `guake.utils` https://github.com/Guake/guake/blob/9c213270767687b30108a7206ac4f4bb913385cd/guake/main.py#L39-L43

As a result, on a wayland session, the pointer position used to get window geometry is always (0,0) resulting in the wrong display being used for window placement and sizing.

See #1820

### DIFF
--- a/guake/main.py
+++ b/guake/main.py
@@ -36,6 +36,9 @@ from optparse import OptionParser
 
 log = logging.getLogger(__name__)
 
+# Force use X11 backend under wayland before any import of GDK through dependencies
+os.environ["GDK_BACKEND"] = "x11"
+
 from guake.globals import NAME
 from guake.globals import bindtextdomain
 from guake.support import print_support
@@ -62,9 +65,6 @@ def main():
     """
     # Force to xterm-256 colors for compatibility with some old command line programs
     os.environ["TERM"] = "xterm-256color"
-
-    # Force use X11 backend underwayland
-    os.environ["GDK_BACKEND"] = "x11"
 
     # do not use version keywords here, pbr might be slow to find the version of Guake module
     parser = OptionParser()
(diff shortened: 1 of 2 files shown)


bitscher:
Note that it could be the cause of other sizing/position-related issues that happen on Wayland.

bitscher:
Also worth noting that the following comment may be incorrect/misleading:
https://github.com/Guake/guake/blob/9c213270767687b30108a7206ac4f4bb913385cd/guake/main.py#L621-L626

zeusintuivo:
After trying this. I think it does something funny. It seems it disappears the icon on "ALT TAB" and when using "Window List" gnome extension, the icon tab is gone. Which leads to the question, Is this how Guake is supposed to work? 
Given that is guake is "a dropdown terminal." it seems that this change does fix some issues. 
Which bring the question, could we make an option of adding a check mark to not hide the from "ALT TAB", and not hide form "Window List" 
At least, I find it useful to have an icon listed and not hid, given that in Red Hat flavoured systems, the "SHOW=HIDE" key combination is not linking any combinations.

eeshugerman:
> It seems it disappears the icon on "ALT TAB" and when using "Window List" gnome extension, the icon tab is gone. Which leads to the question, Is this how Guake is supposed to work?

Yes, I believe so -- the Guake icon should not appear in the app switcher, or GNOME overview, etc. That's how it works under X anyway. In fact, that's how I ended up here: I was about to submit a bug report for this.

Instead I guess I'll just watch this PR and in the meantime override my .desktop file to launch Guake with GDK_BACKEND=x11 :smile:

gsemet:
thanks

PR #1840, opened 2021-04-06
On runtime, require setuptools, not the entire pbr

hroncok:
Hello. I've noticed that guake/paths.py imports from pkg_resources, hence setuptools is required **on runtime**.
When I tried to understand where are the runtime requirements defined, I've noticed a weird to me "Pipfile -> requires.txt -> pbr" chain and I have no idea if I am doing it right.
I've also noticed pbr is required on runtime, which I believe should not be the case.

I have not completed any steps described in your PR template, becase I have just clicked on the GitHub edit button and it does not let me do any of the stuff described here. I am sorry about that, consider this a issue report with a possible fix attached.

Feel free to rework this entirely or just reject it, if that is an issue, however I believe this is the right thing to do.

Thanks for considering.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -15,6 +15,7 @@ typed-ast = "*"
 astroid = "*"
 fiximports = ">=0.1.18"
 mock = ">=2.0.0"
+pbr = "*"
 pytest-mock = "*"
 pyflakes = "*"
 pytest = "*"
@@ -35,5 +36,5 @@ black = "==19.10b0"
 dephell = "*"
 
 [packages]
-pbr = "*"
+setuptools = "*"
 typing = {markers = "python_version < '3.5'"}


gsemet:
no sadly, pbr is a dependency for the whole module handling. It is not a big dependency, but it is needed for execution (version discovery, ...).

the good solution would be to switch to poetry and pyproject.toml that would be a much more cool and modern build system.

hroncok:
So, only adding `setuptools = "*"` is the way to go here?

gsemet:
I do not think so, pbr is a build requirements AND a runtime requirement.

gsemet:
See https://github.com/Guake/guake/blob/master/guake/__init__.py#L28

hroncok:
Yeah, keeping pbr is fine, I've meant *also* adding setuptools.

gsemet:
actually if you look at the code the pkg_resource is not used, we can remove these imports

hroncok:
Even better.

hroncok:
https://github.com/Guake/guake/pull/1843

PR #1843, opened 2021-04-12
Remove unused imports to avoid runtime dependency on setuptools

hroncok:
See https://github.com/Guake/guake/pull/1840#issuecomment-816134048


Davidy22:
Closing this because it is a subset of #1862 which is underway

PR #1846, opened 2021-04-24
Replace Markdown syntax with RST syntax in Fedora section

J3RN:
The rich diff speaks for itself:
image


Jmennius (approved):

Davidy22:
Looking at the file, it is weird how the fedora section is different from everything else. Merged.

Davidy22 (approved):

PR #1862, opened 2021-07-01
Cleanup and enable highly useful flake8 checks

eli-schwartz:
Unused imports and code used without importing it, is quite useful to know about and fix. Stop ignoring those checks, and fix all current issues.

```
 23 files changed, 17 insertions(+), 142 deletions(-)
```

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,5 +1,6 @@
 
 [flake8]
+builtins = _
 ignore =
     E902,
     E226,
--- a/.flake8
+++ b/.flake8
@@ -7,7 +7,6 @@ ignore =
     E302,
     E41,
     E402,
-    F401,
     C901,
     E722,
     W503, # line break before binary operator
--- a/.flake8
+++ b/.flake8
@@ -7,7 +7,6 @@ ignore =
     E41,
     E402,
     F401,
-    F821,
     C901,
     E722,
     W503, # line break before binary operator
--- a/guake/about.py
+++ b/guake/about.py
@@ -29,7 +29,6 @@ from guake import guake_version
 from guake.common import gladefile
 from guake.common import pixmapfile
 from guake.simplegladeapp import SimpleGladeApp
-from locale import gettext as _
 
 
 class AboutDialog(SimpleGladeApp):
--- a/guake/about.py
+++ b/guake/about.py
@@ -22,7 +22,6 @@ Boston, MA 02110-1301 USA
 import gi
 
 gi.require_version("Gtk", "3.0")
-from gi.repository import GObject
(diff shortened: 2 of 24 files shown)


Davidy22 (changes requested):
Hello,

The changes here I assume are flake detect dead code and imports, so I'm alright with folding this in. We've done some updates to fix our CI since this patch was created. so I just need you to do an update and make sure your changes pass CI. Would be nice if you could squash similar commits down maybe.

Davidy22:
I'm going to drop the gettext commit because it fails test cases, but you can recreate the change as a new pull request in the future in a patch that passes the build.

eli-schwartz:
Why do you review my PR and ask for changes, then immediately go ahead and push your own changes to my PR which try... and fail... to do so?

Your change broke the CI on my gettext patch and there is an obvious fix, and now you want to drop that change without even waiting a couple hours for me to show up?

Look, I know you're probably impatient to merge changes which people have contributed months ago without response. But you could give people a couple hours to reply without immediately concluding that that they've lost interest and won't return to address your review comments, then force pushing to delete commits that you are apparently meh about.

Keeping in mind also that it's rather annoying to actually re-aquire a copy of the commit after you've deleted it from my branch and github doesn't expose it visibly past a force push, if I don't have it locally backed up...

Perhaps you're trying to *make* me lose interest and disappear never to return?

I normally leave "allow edits by maintainers" checked as a good faith gesture so that maintainers can do things like rebase or add additional commits, this is the first time I've ever had someone delete them before. Perhaps I should rethink that policy.

Davidy22:
Oh whoops, went ahead with going through and dealing with the things that didn't pass checks myself a day after the initial review. I think I can restore the gettext change for fixing, should still be in my reflog.

Davidy22:
Still needs amending though, the test suite doesn't pick up on the builtin when checking the prefs file, after this there's a recent black update that needs a rebase to check.

eli-schwartz:
Thank you.

One of your CI fixes involved adding another locale.gettext import, which meant that it was being imported globally (your addition) and in `__main__` (my addition) and then raised a lint error for double import.

The remaining issues were black, plus the test suites not running with gettext patched into builtins. This would be a bit simpler if I wanted to import locale and patch builtins in `guake/__init__.py` but I refrained from doing so because this would add a new import to `guake/guake_toggle.py` which is supposed to be as fast as possible, so I instead added it at every script entry point.

eli-schwartz:
Alright, CI passed. \o/

Davidy22 (approved):
Alright, changes look fine and it passes tests, merging.

Issue #924, opened 2017-03-24
show tab bar only when it's needed


deathtrip:
Right now, you can only have the tab bar visible at all times or completely hidden.
Whats needed is the ability to display it only when the user opens a new tab, and auto-hide it when there's only one tab left.

Stibbons:
i like the idea.

Stibbons:
If you want this feature, please vote on [FeatHub](https://feathub.com/Guake/guake)

PR #1867, opened 2021-07-26
Make guake use the system default theme and hide tab bar when there is only one tab

Davidy22:
Detect system theme by default like other gtk programs. Some misc setup/cleanup changes in the fork as well.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -12,5 +12,5 @@ ignore =
     W503, # line break before binary operator
     E231, # missing whitespace after ','
 max-line-length = 100
-exclude = env
+exclude = env,.direnv,docs
 max-complexity = 25
--- a/.gitignore
+++ b/.gitignore
@@ -73,6 +73,8 @@ docs/_build/
 AUTHORS
 ChangeLog
 
+.direnv
+.envrc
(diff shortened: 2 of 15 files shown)


Davidy22:
I should have made a seperate branch for this, but I also did a little hacking and got a thing that should resolve #924 as well.

jacksongoode:
Yes! I was just about to propose hiding with a single tab - nice.

Davidy22:
Hiding with one tab was a resolution to an existing issue report, so someone else beat you to the punch by some amount already.

mlouielu:
Hi @Davidy22, please help to change a branch from your `master` branch to another name, otherwise it may confuse to others.

mlouielu:
We can separate this PR into two (or three), as we would like to have one purpose in one PR:

1. Add option to hide tab bar when we have only one tab
2. Make Guake use system default theme
3. Cleanup some code.

Davidy22:
Alright, just gonna do some name shuffling and push some commits around. Been a while since I wrote this, but I think the cleanup was somewhat related to the themes feature

PR #1874, opened 2021-08-31
feat:Make guake use the system default theme

Davidy22:
Detect system theme by default like other gtk programs. Some misc setup/cleanup changes in the fork as well.

PR #1880, opened 2021-09-07
Update `black` to 21.8b0 and re-format code base

mlouielu:
Update `black` to `21.8b0`  and re-format code base by black using:

```
$ make style
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,7 @@ dev-no-pipenv: clean
 	. .venv/bin/activate && pip install -r requirements.txt -r requirements-dev.txt -e .
 
 pipenv-install-dev:
-	pipenv install --dev --python $(PYTHON_INTERPRETER)
+	PIPENV_IGNORE_VIRTUALENVS=1 pipenv install --dev --python $(PYTHON_INTERPRETER)
 
 ln-venv:
 	# use that to configure a symbolic link to the virtualenv in .venv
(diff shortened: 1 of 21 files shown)


mlouielu:
As @gsemet mention in https://github.com/Guake/guake/pull/1874#issuecomment-914160668, I will use `make style` to do it again.

mlouielu:
hmm, `make style` and `python -m black guake/*.py` gave different result...

gsemet:
yes, update black, and ensure make style does isort, black and all reformatitng stuff.

Davidy22 (commented):
I used `make black` to make my black run work but `make style` is just an alias to `make black` so it'll do the same job.

If we're updating black, you'll want to bump the version number in Pipfile and regenerate Pipfile.lock. I notice that you probably ran the newer version of black on your local machine while the version of black listed in the Pipfile seems unchanged, which is probably why the tests are throwing errors on all the changed files

```
 would reformat /home/runner/work/guake/guake/guake/__init__.py
would reformat /home/runner/work/guake/guake/guake/dialogs.py
would reformat /home/runner/work/guake/guake/guake/globals.py
would reformat /home/runner/work/guake/guake/guake/boxes.py
would reformat /home/runner/work/guake/guake/guake/guake_logging.py
would reformat /home/runner/work/guake/guake/guake/keybindings.py
would reformat /home/runner/work/guake/guake/guake/guake_app.py
would reformat /home/runner/work/guake/guake/guake/notebook.py
would reformat /home/runner/work/guake/guake/guake/settings.py
would reformat /home/runner/work/guake/guake/guake/terminal.py
would reformat /home/runner/work/guake/guake/guake/theme.py
would reformat /home/runner/work/guake/guake/guake/prefs.py
would reformat /home/runner/work/guake/guake/guake/utils.py
Oh no! 💥 💔 💥
13 files would be reformatted, 20 files would be left unchanged.
make: *** [Makefile:192: black-check] Error 1
Error: Process completed with exit code 2.
```

gsemet:
indeed, the version of black should be hardcoded. 21.8b0 is fine but it only support python 3.6.2+

mlouielu:
Weird, black pass test in 3.6 in their GA, but we encounter `dataclasses not found` in 3.6.

Davidy22:
I'm looking online and it looks like dataclasses was added in python 3.7. There is a dataclasses module on pypi available that is an exact clone of the built in python dataclasses module, which we can add to our pipfile to make 3.6 work, or we can drop the five year old 3.6 from our test range.

mlouielu:
Done, please @Davidy22 help to check it.

Davidy22 (approved):
Ran it and it works fine, I notice we went with the dataclasses in pipenv route so just need to remember to drop that dep in the distant future when we drop 3.6

gsemet:
it is ok to use `dataclasses` for python 3.6 support. 3.5 is end of life, so not supported. 3.6 is a good version (first one where i really started enjoying coding in Python!!!).

PR #1883, opened 2021-09-09
Metadata simplify

eli-schwartz:
Use more idiomatic setup.cfg where possible, preferring setuptools built-in features where possible instead of pbr-specific ones.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,7 @@ install-guake:
 	@echo "Please prefer you application package manager (apt, yum, ...)"
 	@echo
 	@echo "#############################################################"
-	@$(PYTHON_INTERPRETER) -m pip install -r requirements.txt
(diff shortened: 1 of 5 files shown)


Davidy22 (approved):
Builds and runs fine, did some searching to make sure that cutting /data would have no side effects and grep doesn't turn up relevant references and the wheel looks fine too. The other changes look like updates to equivalent actions and ci still works, so I'll merge.

eli-schwartz:
Double checking history it seems that the data directory got deleted/moved in 4153ef38f9044cbed6494075fce80acd5809df2b at which point the data_files glob got forgotten.

eli-schwartz:
At this point there is very little pbr still in use, so I am hopeful I will be able to get a modern and clean pure setup.cfg setup in the near future...

PR #1891, opened 2021-09-10
Fix typo, grammar, and doc for fedora install

ersen0:


docs/source/user/installing.rst
```diff
@@ -15,12 +15,12 @@ Ubuntu and Debian users will use ``sudo apt install guake``.
...
 ------
 
-Tested in 29:
+Tested in Fedora 34:
 We can use dnf to install it.
 
 .. code-block:: bash
 
-    $ sudo dnf install -y guake
+    $ sudo dnf install guake
```
Davidy22, line 23: Haven't used fedora in a long while, but isn't the -y for auto yes? Is there a reason why we don't want that?

docs/source/user/installing.rst
```diff
@@ -34,7 +34,7 @@ You may notice the style may have some error. You can fix this by installing and
 
 .. code-block:: bash
 
-    $ sudo dnf install -y arc-theme gnome-tweaks
+    $ sudo dnf install arc-theme gnome-tweaks
```
Davidy22, line 37: Same as previous comment

Davidy22 (commented):
Some nice grammar fixes all-round, just want to know what's happening with a few of the changes before merging this.

docs/source/user/installing.rst
ersen0: I changed this because I think it is a better practice to let users review what they are installing. For example, packages may have optional dependencies ( called `weak dependencies` in Fedora)  and they may want to opt out of installing them.

Davidy22: eh sure alright, not that big an issue with guake though

Davidy22 (approved):

Issue #1643, opened 2019-09-07
Support of short term focus loss and avoid folding


katoquro:
Hello! 

I had an issue with keyboard shortcuts and "Hide on lose focus" option. I'm using https://github.com/autokey/autokey and it always takes focus for very short period of time and then returns it back. 
I've tried to implement a feature when there is a timeout for the window hiding and focus is checked after timeout. I'm not good at Guake technology stack so I decided make a proof of concept and see how it works. I'm using it for several months and didn't found any major issues. 

The code can be found here 
 https://github.com/katoquro/guake/commit/87dcdbb7c6cd5d9f3376c50a7524471725d3a635 
or the last commit at this branch https://github.com/katoquro/guake/commits/feature/short-term-focus-loss
I can create a pull request if some one knows how to pull this feature as flag into the settings 

Looks like this feature can close these issues 
https://github.com/Guake/guake/issues/937 
https://github.com/Guake/guake/issues/687

Davidy22:
You can create a pull request from your fork and we can take a look and help with getting it polished off for inclusion.

#1831 might be related too.

Davidy22:
Alright, I've checked out and rebased your code and I can notice the small delay, but I haven't been able to observe whether or not this actually fixes the issue because my IME didn't steal focus in the first place, and I can't find a program that only swipes focus for a small moment. For adding items to preferences, I have some [open pull requests](https://github.com/Guake/guake/pull/1874) that [demonstrate adding an item to preferences](https://github.com/Guake/guake/pull/1873) that you can follow a similar path for.

Davidy22:
Update after using this patch for a while, I realised that workspace switch also counts as a short focus grabbing event and while it was novel, I don't want guake to stick around when I switch workspaces. This may need functional reworking, also potentially this bug should be the responsibility of the IME developer since some IMEs (like mine, fcitx) are apparently able to function without taking focus and dropping you out of guake.

katoquro:
I've rebased and pushed my PoC to create PR. 
@Davidy22 Could you explain a bit more about the behavior with workspaces? You mean guake terminal keep open in the new workspace after the switch but you want get it closed?

Davidy22:
[...]

PR #1892, opened 2021-09-11
PoC of functionality for lazy focus loss (to avoid closing on short-term losses)

katoquro:
this PR is example of subj functionality 
issue https://github.com/Guake/guake/issues/1643

I had some issues during running lints (if some one could say what is the "black" command I'll be thankful)
```
Error: the command black could not be found within PATH or Pipfile's [scripts].
make: *** [Makefile:186: black] Error 1
```


Davidy22:
Black is a program for code formatting. Running `pip install black` will install black. We have a nice shortcut for formatting all your code with black, if you run `make style` in the top level it'll auto format your code. Just a little thing to force consistent style.

Davidy22:
Oh, I missed that there was an update to this since I last looked. Looks like the linter has some things to say about your code, I'll copy and paste it here for easier access. There's line numbers and filenames so you can see what lines it wants you to fix.

```
guake/guake_app.py:507:16: W0621: Redefining name 'GLib' from outer scope (line 41) (redefined-outer-name)
guake/guake_app.py:507:16: W0404: Reimport 'GLib' (imported line 41) (reimported)
guake/guake_app.py:507:16: C0415: Import outside toplevel (gi.repository.GLib) (import-outside-toplevel)
guake/guake_app.py:518:12: E1121: Too many positional arguments for function call (too-many-function-args)
```

Davidy22:
Alright, I've gone in and done a quick rebase and cleared up the lint errors. Still not sure about making this the default behavior, but I could come around on making this an option. You're going to need to edit the prefs.glade file in glade, then add some code in prefs.py and make your changes conditional on the new preference.

katoquro:
There was no discussion about defaults or so. It's just proof of concept how I solve this issue and I just build suitable for me version from source because I'm not a UI developer at all. If some one find this solution viable and can create corresponding settings they could find comments like [this](https://github.com/Guake/guake/pull/1892/files#diff-be8a9fc8fe311afaa1cf2a8b932b3b371cd3ad3208c5dc0f2dd4695239d52bacR285) describing where settings should be placed

Davidy22:
Oh I see the line, can probably add the setting then some time.

PR #1895, opened 2021-09-13
Restore COMPILE_SCHEMA option

Davidy22:
Restoring as per review https://github.com/Guake/guake/pull/1893#pullrequestreview-752712271

made a little whoops

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@ bindir = $(exec_prefix)/bin
 PYTHON_SITEDIRS_FOR_PREFIX="env PREFIX=$(PREFIX) $(PYTHON_INTERPRETER) scripts/all-sitedirs-in-prefix.py"
 ROOT_DIR=$(shell pwd)
 DATA_DIR=$(ROOT_DIR)/guake/data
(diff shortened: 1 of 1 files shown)


eli-schwartz:
As mentioned in the linked comment, we can look at this as a fortuitous opportunity to change how this works and just check whether DESTDIR is empty. If it is empty, then post install routines such as triggering a glib schemas rebuild of the compiled schemas cache, need to be re-run. If it is not empty, then the path which the files have just been copied to is not the "real" installation and creating the schemas cache will just conflict with the live system once once the two filesystem trees are merged (e.g. by installing the distro package), hence it should not be created.

This will also serve as documentation for the future, that skipping a recompile is relevant specifically to distros and other people who use DESTDIR.

gsemet:
I let you deal with it. it is a tricky part I do not remember how to make it work on all use cases:
- dev environemt with the virtualenv
- make install (probably never really used)
- rpm/deb packaged (the package maintainer may use the result of what make install does)
- pip install guake in users' local env
- and maybe one day pipx !

PR #1902, opened 2021-09-15
feat:Add support for tabs in keyboard shortcuts

### LINKED ISSUE #152: Allow Ctrl-Tab as an accelerator
Allow Ctrl-Tab as an accelerator   http://guake.org/ticket/247
Reported by:    frabjous    Owned by:   somebody
Priority:   minor   Milestone:  
Component:  guake   Version:    0.4.1
Keywords:   keyboard shortcuts  Cc:  
Description

Thanks for the great software!

I would like to assign Ctrl-Tab and Ctrl-Shift-Tab as my keyboard shortcuts for switching to Next Tab and Previous Tab respectively. However, they are not allowed when I go into Guake preferences, click on Ctrl-PageUp/PageDown and it shows choose new acceleator, it simple beeps at me if I press Ctrl-Tab.

Is this because the shell itself uses Ctrl-Tab?

Davidy22:
Also changed the default keybinding for next/prev tab. Had to rewrite half of the keybindings code to make this work, but the incidental refactor along the way was pretty good too.

Resolves #152

The keybinding rewrite is hefty, testing before merging preferred, I'll keep using it for now, haven't noticed anything off between finishing the code and creating this pr.

### DIFF
--- a/guake/data/org.guake.gschema.xml
+++ b/guake/data/org.guake.gschema.xml
@@ -399,12 +399,12 @@
             <description>Move the current tab to the right</description>
         </key>
         <key name="previous-tab" type="s">
-            <default>'&lt;Control&gt;Page_Up'</default>
+            <default>'&lt;Control&gt;&lt;Shift&gt;Tab'</default>
             <summary>Previous tab.</summary>
             <description>Shows the previous tab.</description>
         </key>
         <key name="next-tab" type="s">
-            <default>'&lt;Control&gt;Page_Down'</default>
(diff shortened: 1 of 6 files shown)


besworks:
I just updated my guake install (previous version was only from back around July) and I think something about this PR may have broken my customized keyboard shortcuts which use the super key. Every time I try one the letter I pressed just gets output to the terminal. Other shortcuts I have which do not use the super key still work, and remapping the broken ones also works. Is it just me or can anybody reproduce this behavior?

PR #1907, opened 2021-09-22
Fix serach box disappear when set background image

mlouielu:
We didn't re-paint `RootTerminalBox.search_revealer`, thus search box
will not show when set background image.

Fix this by re-paint search_revealer.


gsemet:
ok merged

PR #1909, opened 2021-09-22
Add upload to test-pipy and pipy (on tags)

gsemet:


gsemet:
Self-approving the PR to test on master

PR #1911, opened 2021-09-24
fix:Do not set GDK_BACKEND to x11 in terminals.

### LINKED ISSUE #1871: GDK_BACKEND is propagated to the shell in terminal
**Describe the bug**

The `GDK_BACKEND` environment variable set by Guake entrypoint is propagated to terminals and, in turn, to all the child processes.
This forces all GUI apps launched from Guake to use XWayland on Wayland session.

On a Wayland session, setting `GDK_BACKEND` to `x11` while `XDG_SESSION_TYPE` is set to `wayland`, can cause apps misbehaviour.
E.g. GTK-based flatpak apps with `--socket=fallback-x11` permission (which allows x11 socket only on X11 session) will fail to launch with `Gtk-WARNING **: 18:18:53.576: cannot open display:`, unless `GDK_BACKEND` is unset in advance.

#1763 seems somehow related to this.

**Expected behavior**

Guake itself runs with `x11` GDK backend, while `GDK_BACKEND` env var is unset in the shell.

**Actual behavior**

`GDK_BACKEND` env var is set to `x11` in the shell.

**To Reproduce**

1. Launch Guake on Wayland session
2. Run `echo $GDK_BACKEND` in the terminal
3. Observe `x11` in output

-------------------------------------------------------------------------------

<details><summary>$ guake --support</summary>

Guake Version:		3.7.0

Vte Version:		0.64.2

Vte Runtime Version:	0.64.2

--------------------------------------------------
GTK+ Version:		3.24.30

GDK Backend:		GdkX11.X11Display

--------------------------------------------------
Desktop Session: gnome

--------------------------------------------------
Display: :0

RGBA visual: True

Composited: True

* Monitor: 0 -  XWAYLAND0
    * Geometry:		2560 x 1440 at 0, 0
    * Size:		550 x 310 mm²
    * Primary:		True
    * Refresh rate:	59.91 Hz
    * Subpixel layout:	unknown

Davidy22:
Do not set GDK_BACKEND to x11 in terminals.

Don't automatically inherit env variables and manually pass modified env

Fixes  #1871

### DIFF
--- a/guake/terminal.py
+++ b/guake/terminal.py
@@ -121,6 +121,10 @@ class GuakeTerminal(Vte.Terminal):
 
         self.setup_drag_and_drop()
 
+        self.ENVV_EXCLUDE_LIST = ["GDK_BACKEND"]
+        self.envv = [f"{i}={os.environ[i]}" for i in os.environ if i not in self.ENVV_EXCLUDE_LIST]
+        self.envv.append(f"GUAKE_TAB_UUID={self.uuid}")
+
     def setup_drag_and_drop(self):
         self.targets = Gtk.TargetList()
         self.targets.add_uri_targets(DropTargets.URIS)
@@ -550,12 +554,13 @@ class GuakeTerminal(Vte.Terminal):
             Vte.PtyFlags.DEFAULT,
             directory,
             argv,
-            [f"GUAKE_TAB_UUID={self.uuid}"],
-            GLib.SpawnFlags.DO_NOT_REAP_CHILD,
+            self.envv,
+            GLib.SpawnFlags(Vte.SPAWN_NO_PARENT_ENVV | GLib.SpawnFlags.DO_NOT_REAP_CHILD),
             None,
             None,
             None,
         )
+
         try:
             tuple_type = gi._gi.ResultTuple  # pylint: disable=c-extension-no-member
         except:  # pylint: disable=bare-except
--- a/releasenotes/notes/fix_gdk_backend-be5c40f221c1e528.yaml
+++ b/releasenotes/notes/fix_gdk_backend-be5c40f221c1e528.yaml
@@ -1,0 +1,6 @@
+release_summary: >
+    Stopped setting GDK_BACKEND to x11 in terminals.
+
+fixes:
+  - |
+      - GDK_BACKEND is propagated to the shell in terminal #1871 


guake/terminal.py
```diff
@@ -121,6 +121,10 @@ def __init__(self, guake):
 
         self.setup_drag_and_drop()
 
+        exclude = {"GDK_BACKEND"}
```
mlouielu, line 124: we may put this as a constant?

```
self.ENVV_EXCLUDE_LIST = ["GDK_BACKEND"]
```

guake/terminal.py
```diff
@@ -121,6 +121,10 @@ def __init__(self, guake):
...
+        self.envv = [f"{i}={os.environ[i]}" for i in os.environ if i not in exclude]
```
mlouielu, line 125: I believe we cannot do it at new terminal spawn.

User may add/append new environment variable after they spawn the terminal,

e.g.

```
#
# after spawn new terminal
#

$ export FOO=bar

# Spawn new termianl

$ echo $FOO  # This won't show anything, because we didn't spawn with this env var

$
```

guake/terminal.py
```diff
@@ -550,12 +554,13 @@ def spawn_sync_pid(self, directory):
             Vte.PtyFlags.DEFAULT,
             directory,
             argv,
-            [f"GUAKE_TAB_UUID={self.uuid}"],
-            GLib.SpawnFlags.DO_NOT_REAP_CHILD,
+            self.envv,
```
mlouielu, line 557: A better place is to maintain envv at this point each time we need to spawn new termianl.

mlouielu (changes requested):

guake/terminal.py
Davidy22: Setting environment variables in the new terminal still works upon testing. Unless you mean setting new environment variables in the parent and having them auto propagate down to already running children? In which case I don't think that was happening already anyway.

guake/terminal.py
Davidy22: Alright, sure, could be useful at some point.

guake/terminal.py
Davidy22: That was my first draft, but then I noticed that the value of self.envv was the same every time so I moved the generation up to __init__ to do it once.

guake/terminal.py
mlouielu: At first glance I though export variable will be set in spawn terminal, but it seems I'm wrong.

Could I sync with you the test step?

```
1. in terminal type `export FOO=bar`
2. right click & `split |` to spawn terminal
3. echo $FOO will show nothing in spawn terminal
```

Davidy22: This is what happens yeah. The bug is caused by environment variables from the place that guake is initially run from being propagated to all spawned terminals, and in particular one line in main.py at line 45 that sets GDK_BACKEND in the code. Terminal splitting shouldn't have anything to do with the bug.

mlouielu: Got it, thanks for you explain

mlouielu (changes requested):
Function test pass, LGTM.

Only a small point, please help remove duplicate commit title in commit message.

Davidy22:
Done.

mlouielu (approved):
LGTM, but I forgot one thing (I thought it is like GitLab...), please also put `Fixed: #1871` in commit message!

Then it is good to merge.

Davidy22:
eh alright done

mlouielu (approved):
Thanks, LGTM

mlouielu:
I just find out that merge commit will be editable after I click merge..., so basically my last request can be added in merge commit next time :100:

PR #1912, opened 2021-09-28
Fix ci

Davidy22:
Fix the redownload path, fixed the upload part so that it actually only uploads for one of the python versions, and bump the version that's uploaded to 3.9 because new. Didn't actually check the pip upload portion, only fixed the path issue but if the pip key is good this should fix the build.


gsemet:
thanks. i'll try doing a rc

PR #1920, opened 2021-10-02
Add Finnish translations

jonnelafin:


po/LINGUAS
```diff
@@ -1,2 +1,2 @@
 # Set of available languages.
-ca cs de el es fa fr gl hr hu id it ja ko nb nl pa pl pt_BR ru sv tr uk zh_CN zh_TW
+ca cs de el es fa fr fi_FI gl hr hu id it ja ko nb nl pa pl pt_BR ru sv tr uk zh_CN zh_TW
```
Davidy22, line 2: Why is the country code fi_FI and not just fi? I'm looking at [this resource](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) and I see finnish listed as just fi, like some number of other languages are just two letters. The preexisting exceptions look to be when it's a language with regional differences, like zh_CN and zh_TW. Is there a need for that with finnish?

Davidy22 (commented):
Looks good, needs a release notes file but I'll write that. In the meantime, just want to ask something about the country code

po/LINGUAS
jonnelafin: fi_FI seems to be incorrect indeed. I had always presumed it had to do with Finland officially being a multilangual country (Finnish and Swedish). There seems to be some cases of fi_FI being used incorrectly in other projects too: [nextcloud](https://github.com/nextcloud/appstore/issues/479) and [woocommerence](https://github.com/woocommerce/woocommerce/issues/6302). 

I have amended the commit to fix this.

Davidy22: Alright, I'll add a release notes file and if I haven't messed up the formatting on the file and all checks pass I'll approve and merge.

Davidy22:
Passes checks, merging.

Davidy22 (approved):

PR #1921, opened 2021-10-04
News file update

Davidy22:
Prep for release
Kind of want to see if there's a way to disable regenerating older news entries

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -2,6 +2,186 @@
 guake
 =====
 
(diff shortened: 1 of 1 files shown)


mlouielu (approved):
LGTM, this is auto generated right?

Davidy22:
I did edit it slightly, the fully auto generated version has some lines that say `- None` from before I knew that those would show up in the auto generated notes.

Issue #123, opened 2013-04-30
localization issues


kikecalpe:
localization issues  http://guake.org/ticket/291
Reported by:    ulidtko     Owned by:   somebody
Priority:   major   Milestone:  0.5.0
Component:  translations    Version:  
Keywords:       Cc:  
Description

```
"selectd" typo on [source:/src/guake#L1283].
Unclear "(The trayicon was enabled)" on [source:/src/guake#L635]. What does it mean? Is it necessary there? Can it be removed?
Does GPL need to be translated? 
```

PR #1928, opened 2021-10-10
Fix metainfo directory prefix

sibskull:
Commit c9a7c66 placed guake.desktop.metainfo.xml to $(DESTDIR)/metainfo/
instead of $(DESTDIR)$(datadir)/metainfo/. In package this file will be
placed to /metainfo, but standard
(https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#qsr-app-introduction)
says:

"Applications can ship one or more files in /usr/share/metainfo/%{id}.metainfo.xml"

Return $(datadir) to installation path for metainfo file.

- rebase on latest HEAD:

  ```bash
  ```

- hack your change

  ```bash
  ```

  generation, using:

  ```bash
  ```

- create new commit message

  ```bash
  $ <hack the code>
  $ git commit --all
  ```

  line in your commit messages:
  
  ```
  # For a bug fix, uses:
  sem-ver: bugfix
  
  # For a new feature, uses:
  sem-ver: feature
  
  ```

### DIFF
--- a/releasenotes/notes/move_to_datadir-6d95d29ded5fa040.yaml
+++ b/releasenotes/notes/move_to_datadir-6d95d29ded5fa040.yaml
@@ -1,0 +1,6 @@
+release_summary: >
+    Moved guake.desktop.metainfo.xml to $datadir
+
+notes_for_package_maintainers:
+  - |
+    guake.desktop.metainfo.xml has been moved to $datadir in Makefile


chax:
> On the point of consolidating more lines to $datadir, if those lines idiomatically belong to $datadir, if they should change dir if datadir changes, we can swap them as well.

Well, i think it makes sense to fix also lines 135 and 136 because `${datadir}` is also used in uninstall-schemas target of Makefile
https://github.com/Guake/guake/blob/a8283040e0d1d068043f6850ead27796710789df/Makefile#L135-L136
https://github.com/Guake/guake/blob/a8283040e0d1d068043f6850ead27796710789df/Makefile#L164

PR #1935, opened 2021-10-18
added is_visible returning boolean Guake show/hide status

argoyal:
Added is_visible flag that returns the value of hidden variable from the guake instance which is unset and set on show and hide functions respectively.

### DIFF
--- a/guake/main.py
+++ b/guake/main.py
@@ -362,6 +362,13 @@ def main():
         help=_("Show support information"),
     )
 
+    parser.add_option(
+        "--is-visible",
+        dest="is_visible",
+        action="store_true",
+        default=False,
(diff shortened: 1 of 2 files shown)


mlouielu:
Hi @argoyal, could you tell us why you close this PR? At first glance it should work as what it should be.

argoyal:
> For the release notes file, you'll want to actually fill it in as well. Delete the sections that aren't relevant, and fill in the sections that are. I'll test this out in the meantime

@Davidy22 do you want me to force push to keep the commit history clean or do you want me to create a separate commit message for these changes.

Davidy22:
This PR has been supplanted by #2008 which works, closing.

PR #1953, opened 2021-10-25
Use keycode for keybindings

### LINKED ISSUE #1946: Keybindings does not work with not English layout.
Keybinds with layout sensitive keys (letters and other) doesn't work.
Test on Arch Linux.

In Guake 3.7.0 keybindings works on both layouts, I think because there used `Gtk.AccelGroup`.
This was changed in commit 6af0a95235a1bf4d6016ef9a3698601e76387fdf and now `keyval` is used.
Maybe we should back to `Accel` realisation with build-in magic.

For example copy keys (`Ctrl+Shift+C`)
For US layout
```
keyval:  67
mod:  <flags GDK_SHIFT_MASK | GDK_CONTROL_MASK | GDK_MOD2_MASK of type Gdk.ModifierType>
scancode/keycode:  54
```
For RU layout
```
keyval:  1779
mod:  <flags GDK_SHIFT_MASK | GDK_CONTROL_MASK | GDK_MOD2_MASK | GDK_MODIFIER_RESERVED_13_MASK of type Gdk.ModifierType>
scancode/keycode:  54
```

-----
As workaround convert `keycode` to english `keyval`, for me it look like this:
```
keybindings.py

def activate(self, window, event):
        key = event.keyval
        mod = event.state
        keycode = event.get_keycode()
        if keycode[0]:
            ent = Gdk.Keymap.get_for_display(window.get_display()).get_entries_for_keycode(keycode[1])
            if ent[0]:
                key = ent[2][0]
        ...
```
This code does not full tested so it can crash in some case.

vantu5z:
Try to fix #1946.

### DIFF
--- a/releasenotes/notes/use_keycode_for_keybindings-a4fb151a046a28c9.yaml
+++ b/releasenotes/notes/use_keycode_for_keybindings-a4fb151a046a28c9.yaml
@@ -1,0 +1,5 @@
+release_summary: >
+  Use keycodes instead of keyvals for keybindings. This allow to use keybindings with different keyboard layouts.
+
+known_issues:
+  - keybindings does not work with not English layout. #1946


fx-kirin:
This change breaks shortcut with custom key mapping through xkb. It would be nicer there is an option to choose the behavior.

This https://github.com/Guake/guake/issues/2058 might be a case too.

PR #1956, opened 2021-10-26
Fix release process

gsemet:
###


Davidy22:
Some futile attempts to fix this [here](https://github.com/Davidy22/guake/tree/fixci), amendments to work towards making this work welcome.

Davidy22:
After far too much wasted time in my branch, I went nuclear and removed everything except the one thing that's actually in the title of this PR and it worked. I'm going to close this PR and create a new one just for the change to turn on the pypi build. For the translations you can remake that as a seperate PR because there's probably a bunch of new stuff on weblate since this was created, and the other changes contain whatever it was that made this so hard to fix so I'm not going to include those in my PR.

gsemet:
It is possible that it would be easier to simply rewrite all the build and release process.

Davidy22:
I mean, the thing I had where I just take all the stuff that isn't pypi release out of this PR works fine

PR #1961, opened 2021-10-29
Add instructions for PPA install

Davidy22:
As per #1960, adding PPA install instructions.


mlouielu (commented):
Add comment in discussion: https://github.com/Guake/guake/discussions/1960

Issue #1934, opened 2021-10-17
Can't make Guake run under Wayland


VitalyAnkh:
**Describe the bug**
After updating Guake to 3.8.0, I found it running under x11 rather than wayland. My desktop is Gnome 41 on Wayland, and before 3.8.0 Guake runs under Wayland as I expect. And I didn't set `GDK_BACKEND` environment variable to not break electron apps.

**Expected behavior**
Guake runs under Wayland

**Actual behavior**
Guake doesn't run under Wayland

**To Reproduce**
Use Gnome 41 on Wayland, install guake 3.8.0, and launch it.

-------------------------------------------------------------------------------

Please run `$ guake --support`, and paste the results here. Don't put backticks (`` ` ``) around it! The output already contains Markdown formatting. And make sure you run the command **OUTSIDE** the Guake.
<details><summary>$ guake --support</summary>

Guake Version:		3.8.0

Vte Version:		0.66.0

Vte Runtime Version:	0.66.0

--------------------------------------------------
GTK+ Version:		3.24.30

GDK Backend:		<GdkX11.X11Display

--------------------------------------------------
Desktop Session: gnome

--------------------------------------------------
Display: :0

RGBA visual: True

Composited: True

* Monitor: 0 - XWAYLAND0
    * Geometry:		1920 x 1080 % 2 at 0, 0
    * Size:		600 x 330 mm²
    * Primary:		True
    * Refresh rate:	59.980000000000004 Hz
    * Subpixel layout:	unknown

VitalyAnkh:
I think Wayland is the future of Linux desktops. We should make Guake run under Wayland by default.

ccolorado:
@Davidy22  I am read the first line of your statement and can't figure out what this means. 
`don't return when removing it. ` 
I can probably try building from source.

fancypantalons:
As an FYI, on my Gnome 41 system, issue #1817 resurfaces when using the Wayland backend.

VitalyAnkh:
@fancypantalons I also meet #1817 if guake runs under Wayland. But if you want Guake to run under x11, you could set `GDK_BACKEND=x11` to guake’s launch command by yourself. Hardcoding this environment variable is bad.

Davidy22:
[...]

PR #1963, opened 2021-10-31
Remove GDK_BACKEND to force x11

Davidy22:
After much feedback from the wayland users who don't want Guake to run in xwayland

Created in response to much feedback from #1934 and other related issues

### DIFF
--- a/guake/main.py
+++ b/guake/main.py
@@ -41,9 +41,6 @@ from optparse import OptionParser
 
 log = logging.getLogger(__name__)
 
-# Force use X11 backend under wayland before any import of GDK through dependencies
-os.environ["GDK_BACKEND"] = "x11"
-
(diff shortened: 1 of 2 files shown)


VitalyAnkh:
Could we merge this?

mlouielu:
> Could we merge this?

Let me take some time on this

It seems more suitable with a revert, but I cannot access to my laptop now.

mlouielu (changes requested):
I would like to have another discussion in #1934, which I proposed not to remove force x11 backend as default, but add `GUAKE_ENABLE_WAYLAND=1` as an alternative way to use `GDK_BACKEND=wayland`.

Davidy22:
Closing since this'll be supplanted with an option

Issue #1845, opened 2021-04-20
Minimal Right Click Menu, no copy or split screen


roizcorp:
I reinstalled my OS, and the Right Click menu never had the additional actions. The weird thing it has used to be on it prior to the OS reninstall, and I do not see any reason why it is missing, the functions do work with keyboard shortcuts e.g. split screen 

```
<details><summary>$ guake --support</summary>

Guake Version:		3.7.0

Vte Version:		0.64.0

Vte Runtime Version:	0.64.0

--------------------------------------------------
GTK+ Version:		3.24.28

GDK Backend:		GdkX11.X11Display

--------------------------------------------------
Desktop Session: gnome

--------------------------------------------------
Display: :0

RGBA visual: True

Composited: True

* Monitor: 0 -  XWAYLAND0
    * Geometry:		1920 x 1080 at 0, 0
    * Size:		280 x 160 mm²
    * Primary:		True
    * Refresh rate:	59.96 Hz
    * Subpixel layout:	unknown
```

Davidy22:
The right click menu is populated from a combination of a custom commands json file and a set of defaults, although it seems like the defaults don't load when the custom commands file is invalid. To fix this, you can create a custom_commands.json file with the contents:

```
{}
```

And point "custom command file path" in the general tab of preferences to your custom commands file to get back all your right click menu options. For a more long term solution we should still have the default command options populate when the custom commands file is invalid.

roizcorp:
hi,
I fif

> The right click menu is populated from a combination of a custom commands json file and a set of defaults, although it seems like the defaults don't load when the custom commands file is invalid. To fix this, you can create a custom_commands.json file with the contents:
> 
> ```
> {}
> ```
> 
> And point "custom command file path" in the general tab of preferences to your custom commands file to get back all your right click menu options. For a more long term solution we should still have the default command options populate when the custom commands file is invalid.

I did as you said, it does not work, no split screen nor copy/paste menu items still
many thanks

Davidy22:
Attaching a valid custom commands file should work around this for older versions, but there's also a new version of guake out now that should fully resolve this issue.

roizcorp:
Can you post it as a reply in a code block, so I could copy and paste it instead?

Davidy22:
{} should have been sufficient on older versions so I suspect there was an issue with file paths, but you can also upgrade to the newest guake where this is fully fixed.

roizcorp:
```
local/guake 3.8.0-1
[...]

PR #1966, opened 2021-11-02
Replace ｜ with non unicode dependent |

Davidy22:
To make vertical bar render on less character encoding proficient machines

Resolving the part of #1845 that doesn't need any maybe questionable working around.


mlouielu (approved):
LGTM

Issue #1826, opened 2021-01-08
Guake tabs start user sessions which don't end when closing tabs. 


Ysgorg:
Each time I open a new tab in Guake, the number of users reported by the `users` command increases. Closing tabs does not decrease the number. Example, for user named `foo` after opening three guake tabs: 

```sh
users
foo foo foo
```

Workaround: `sudo killall guake`

This is unlike the behavior of the built-in terminal, where the output of `users` is unaffected by the number of tabs. 

```sh
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.1 LTS
Release:	20.04
Codename:	focal

$ gnome-shell --version
GNOME Shell 3.36.4

$ guake --version
Guake Terminal: 3.6.3
VTE: 0.60.3
VTE runtime: 0.60.3
Gtk: 3.24.20

```

Davidy22:
Was able to recreate the user creation trivially, and the users not being cleaned up occurs when closing tabs with CTRL+D but not with the close tab keybinding.

Issue #1950, opened 2021-10-25
Core Dump When Starting Guake 3.8.0


LunaGNUisance:
When launching Guake, it will load for 20+ seconds, and nothing will happen. Upon trying to open Guake again, the screen freezes and GNOME core dumps for a few seconds, and then the application opens. This only happens on Guake 3.8.0 and not the previous version.

**Expected behavior**

When opening Guake, it opens without GNOME core dumping.

**Actual behavior**

Guake takes a long time to open and only opens after GNOME core dumps.

**To Reproduce**

Open Guake on GNOME. I was able to reproduce this behavior on three different Arch Linux GNOME machines. On the third machine, something slightly different happens; it kicks me out to the display manager, and I have to login again, and then try to open Guake again. It may or may not work the second or third time.

**Other Details**

I have Guake set to autostart on login, and I have `guake-toggle` set to a keybinding that I use to bring up Guake. Looking at journalctl, there are many lines of "Could not resolve keysm" after GNOME core dumps.

Guake behaves as normal after managing to get into the application. Additionally, on two of the machines with the exact same behavior, I have the AppIndicator extension installed, and when launching Guake, the Guake icon in the system tray is black until it core dumps. Not sure how relevant that is, but it's consistent behavior.

-------------------------------------------------------------------------------

<details><summary>$ guake --support</summary>

Guake Version:		3.8.0

Vte Version:		0.66.0

Vte Runtime Version:	0.66.0

--------------------------------------------------
GTK+ Version:		3.24.30

GDK Backend:		<GdkX11.X11Display

--------------------------------------------------
Desktop Session: gnome-xorg

--------------------------------------------------
Display: :1

RGBA visual: True

Composited: True

* Monitor: 0 - NCP eDP1
    * Geometry:		1920 x 1080 at 0, 0
    * Size:		340 x 190 mm²
    * Primary:		True
    * Refresh rate:	120.035 Hz
    * Subpixel layout:	unknown

LunaGNUisance:
I figured out what the cause was.

I have a variant of this added as a hook for Guake on all of my GNOME computers:

```
guake --execute=cu
```

I have `cu` set as a bash alias for `checkupdates`, which is a script that is part of `pacman-contrib` which will check my mirrorlist for package updates. I have this set on all of my GNOME machines.

I removed it for all 3 computers and it no longer core dumps on startup. Sorry; I really should have figured that out sooner...

vantu5z:
[...]

PR #1974, opened 2021-11-16
Fix vte spawn_sync runtime check failed

mlouielu:
When running Guake, `spawn_sync` shows runtime check failed:

```
$ guake
...
(guake:193534): VTE-WARNING **: 10:57:38.556: (../vte/src/vtepty.cc:667):bool
_vte_pty_spawn_sync(VtePty*, const char*, const char* const*,
                    const char* const*, GSpawnFlags, GSpawnChildSetupFunc,
                    gpointer, GDestroyNotify, GPid*, int,
                    GCancellable*, GError**):
runtime check failed: ((spawn_flags & ignored_spawn_flags()) == 0)
...
```

This is cause by setting `DO_NOT_REAP_CHILD` flag in it.

As shown in vte document & source code, `DO_NOT_REAP_CHILD` is
the default behavior for `spawn_sync`, so we don't need to pass
it when we try to `spawn_sync`.

### DIFF
--- a/guake/terminal.py
+++ b/guake/terminal.py
@@ -569,7 +569,7 @@ class GuakeTerminal(Vte.Terminal):
             directory,
             argv,
             self.envv,
-            GLib.SpawnFlags(Vte.SPAWN_NO_PARENT_ENVV | GLib.SpawnFlags.DO_NOT_REAP_CHILD),
+            GLib.SpawnFlags(Vte.SPAWN_NO_PARENT_ENVV),
             None,
             None,
             None,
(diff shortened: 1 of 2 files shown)


Davidy22 (approved):
Looks good, code runs but I actually already knew that from when I was adding the SPAWN_NO_PARENT_ENVV flag, I left the flag there out of worry of breaking something that I was unaware of but I guess I didn't have to worry.

In somewhat related matters, I tried to replace spawn_sync with spawn_async but gave up partway because it broke an ungodly amount of things. spawn_sync is deprecated though, and eventually we probably want to make the move

mlouielu:
Lets add migrate to `spawn_async` into backlog?

Davidy22:
It was brought up as an aside in #1950, and I think it's probably the fix for that and some other issues too. I also noticed it incidentally resolved  #1826 too when I was testing it

PR #1990, opened 2021-12-16
Fix CI failure of PR #1989

jcfstabile:
Downgrade pipenv to 2021.5.29

PR #1992, opened 2021-12-17
Revert "Fix module gi not found when make run "

mlouielu:
Reverts Guake/guake#1989

We will revise this in #1990


Davidy22:
We can just fix it in the other PR and skip this revert step, can we not?

Davidy22 (approved):
After investigation into error messages on other probably affected PRs, I see reports that a regression past 2018.11.26 is probably causing these issues so I'll second this and send this back to the drawing board.

PR #2002, opened 2022-01-11
Fix upload to pypi test

Davidy22:
After much discouragement from the last several unfruitful attempts to make pypi happen, I fell off a little bit and was busy elsewhere but found it in me recently to get over and take a look at the repo, which had a failing CI build that I somehow failed to notice for ages. Looking at the specifics of the failure, this seems actually encouragingly like a working pypi build is finally in reach, just need to strip out some elements from the version number that pypi can't handle.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,7 +66,7 @@ jobs:
         with:
           a: ${{ matrix.python-version }}
           b: 3.9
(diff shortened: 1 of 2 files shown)


Davidy22:
Merging to get this in because it's pretty important and I'd have to wait weeks otherwise

Issue #1926, opened 2021-10-08
Add an --is-visible option


lgarczyn:
Suggestion:

add a `guake --is-visible` option

Would simply return 1 if guake is currently visible, and 0 if it is not

Would allow third party software to have much more complex interactions with guake

Davidy22:
Completed in #2008

lgarczyn:
Hey, thanks!

PR #2008, opened 2022-01-15
feature: --is-visible option #1926

rjeb:
Implements the --is-visible option suggested in #1926. Screenshots are a from the build on a clean install of Fedora. 
True:
True
False: 
False


Davidy22:
I've just clicked approve for the non-deepsource CI items, and it looks like there's some things that deepsource wants to highlight for fixing as well. I'll take a deeper look after the CI moans are dealt with

rjeb:
Yeah I think the mergeable-if-statement flags have been around for a bit since they appear in this [PR](https://github.com/Guake/guake/pull/1972/checks) from back in November. The unnecessary comment flags seem to be new though.

Davidy22 (approved):
Hum, went in and looked and it does look the CI moans don't have anything to do with your code, not sure why deepsource has decided to do this or how these didn't get raised earlier, but I'll go in and clean that up later. For now, took a look at the code, looks fine and works as expected so I'll merge.

PR #2013, opened 2022-01-18
Update translation from Weblate

gsemet:


Davidy22:
Can we not add the additional ci-breaking commits that don't have anything to do with translation?

gsemet:
yes I have removed it

gsemet:
@Davidy22 do you approve?

NEWS.rst
```diff
@@ -8,18 +8,11 @@ guake
...
 
-Fix system font application issue
-
-Add fallback for version number finding
-
-Deprecated pbr
-
-Use keycodes instead of keyvals for keybindings. This allow to use keybindings with different keyboard layouts.
-
+This new version of Guake tries to catch up with
```
Davidy22, line 11: Why is most of the release summary section deleted?

Davidy22 (changes requested):

NEWS.rst
gsemet: ok i have fixed this glitch

NEWS.rst
```diff
@@ -8,18 +8,11 @@ guake
...
 New Features
 ------------
 
-- --is-visible option returns 1 when visible, and 0 when not
+- ``--is-visible`` option returns 1 when visible, and 0 when not
```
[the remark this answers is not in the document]
gsemet: fixed

Davidy22:
???????????????????????

You deleted more of the release summary section, then force pushed, then made a release? That the opposite of fixing the things I brought up, that's even worse. Can you at least check the file output yourself before you do these things? What command are you using to do this, because this is completely different from the output I get from `make release-note`, after fixing the YAML error. For one, why did you go back and edit an existing reno file that had nothing to do with this PR? And why did you edit them into invalid YAML that breaks `make release-note`? That change made CI fail, that should have been the first red flag to at least open another PR to fix it.

Also, is the readthedocs build getting fixed any time soon?

gsemet:
Sorry I wanted to advance the release. The CI is ok actually I just killed a pipeline, but I don’t know why there is 2 pipelines per MR, one fails, one succeeded.
The RC package is on Pipy so can be tested.

For the Reno files, you can actually edit slugs from anywhere to fix typos and mistakes from previous releases, it is a feature of it. I do not think i have broken anything, if so this can be fixed in a new commit.
There is a lot of maintainance to do on this release process, it is outdated, still use pipenv and setuptools that prevent updating some package for some weird reason.

I do not have time to do everything (baby nbr 2 is arriving this month…), so I go a bit fast sorry.

there is still the readthedoc to fix, indeed.

gsemet:
If you look at this CI it succeeded and uploaded the pipy package is https://github.com/Guake/guake/actions/runs/1714896910

you have reverted my commits too fast. If you agree I will force push back to get a clean history. I do not want to revert your reverts that makes no sense.
Please avoid reverts without at least let me explained what has been done. 

PS: there are not a lot of contribution on Guake, so even if the mainline breaks for a few hours or days it is not the end of the world.

mlouielu:
If possible, we should avoid merging our PR anyway, at least it should be review by others : )

gsemet:
I have seen what you mean, yaml is fixed in https://github.com/Guake/guake/pull/2015. I have fixed the dual pipeline weirdness in https://github.com/Guake/guake/pull/2016

Davidy22:
There is a second pipeline because we added deepsource some time ago as an auxiliary CI process to quality check our code. The one you broke was the one that guake's always had, the one that has our own test cases and build test. If you click on the red cross before you force the PR, it very helpfully tells you what you broke, so that you can fix it before you commit it to the main branch. Passing deepsource and breaking the test cases is not ok. In a similar vein don't cite a workflow you passed when [the same commit very visibly fails another check](https://github.com/Guake/guake/commit/d3fb42faa989c88f16a3a538fb4dd9726fadad01)

This PR is about updating translations. Or at least that's what this PR was supposed to be about. All the breakage to things that have nothing to do with translations in the last few commits shouldn't have happened here, including the modifications to release note files.

I will expedite reverting commits if you expedite merging broken commits. We didn't force push to clean the history the [last time we reverted something breaking that you forced through](https://github.com/Guake/guake/pull/1992) so if that's something we're doing we have two cases to handle here, but preferably our git history matches our PR history.

gsemet:
so for deepsource we can have a simpler and faster pipeline, that won't do what the "push" pipeline does.

PR #2015, opened 2022-01-19
Revert "Revert "doc: update NEWS.rst""

gsemet:
This reverts commit d5c58117595bbf8698d8deea0e303dd951b1635d.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ reset:
 
 all: clean dev style checks dists test docs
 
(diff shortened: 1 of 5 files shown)


gsemet:
@Davidy22 this reverts the revert and fix the yaml issue. And the pipeline is green :D

Davidy22 (changes requested):
The original commit had more wrong with it than just the yaml breakage, by the time we've fixed that this PR isn't a reversion anymore so this needs to be renamed.

Davidy22:
Superceded by #2024

PR #2016, opened 2022-01-19
ci: no pipeline on pull_request

gsemet:


mlouielu:
I believe we still need `pull_request` for other contributor, we can reduce it to `build` with `push, pull_request`, and `release_test` with `push` only.

gsemet:
I do not see the difference between them? If the checks has been done on the push of the branch, why redoing them after each even on the pull request? It would make sense if we would retrieve something from the PL and check it, but we do not do that

Davidy22:
CI on pull request is there so that we can see at a glance if someone is trying to make a pull request that breaks things, like editing old release notes files in a way that breaks release note generation. CI on push catches commits that can happen via avenues other than PR, and lights the badge up. Neither are considerable for removal.

gsemet:
Sorry but I do not agree. If you look in this PL, there is only checks on "push" event, and it does the protection like you say.

"pull_request" only react to event on the pull requests, like comments, review, etc. There is no use here, since we do not consume anything from it. For exemple, on our Gitlab we do that, we retrieve something from the merge request API (checklist status, review,...) and we have a job that can read this information and process it and fail or not.

For guake, we do not use it. Events on pull request does not change the commits. If there is any change, another push event is generate.

Reopening.

Davidy22 (approved):
Oh hum, I always just figured doubled up entries was how it looked. It seems like github is still optimising it down to one run each in the job report, but trimming the display'll be nice.

gsemet:
thx for approval. I merge

PR #2020, opened 2022-01-24
Eliminate redundant terminal spawning

Davidy22:
Panes are always stored even in single paned tabs, the condition gating restoring box layout is always true so the initial terminal spawned was always immediately overwritten

No pull requests I can think of are fixed by this, although the side effects of the duplicate spawning may be the cause of some bugs. Will probably also improve startup times if user has intensive startup scripts or large save sessions

### DIFF
--- a/guake/guake_app.py
+++ b/guake/guake_app.py
@@ -1350,16 +1350,15 @@ class Guake(SimpleGladeApp):
             for index in range(nb.get_n_pages()):
                 try:
                     page = nb.get_nth_page(index)
-                    if page.child:
-                        panes = []
-                        page.save_box_layout(page.child, panes)
-                        tabs.append(
-                            {
-                                "panes": panes,
-                                "label": nb.get_tab_text_index(index),
-                                "custom_label_set": getattr(page, "custom_label_set", False),
-                            }
-                        )
+                    panes = []
+                    page.save_box_layout(page.child, panes)
+                    tabs.append(
+                        {
+                            "panes": panes,
+                            "label": nb.get_tab_text_index(index),
+                            "custom_label_set": getattr(page, "custom_label_set", False),
+                        }
+                    )
                 except FileNotFoundError:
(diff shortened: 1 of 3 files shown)


guake/guake_app.py
```diff
@@ -1350,16 +1350,15 @@ def save_tabs(self, filename="session.json"):
...
-                        panes = []
-                        page.save_box_layout(page.child, panes)
-                        tabs.append(
-                            {
-                                "panes": panes,
-                                "label": nb.get_tab_text_index(index),
-                                "custom_label_set": getattr(page, "custom_label_set", False),
-                            }
-                        )
+                    panes = []
```
mlouielu, line 1353: Will this cause regression for 6cc235430f525c999420504e5b49d77bb3669c25?

mlouielu:
I'll take a closer look at night.

Davidy22:
This PR is pretty solidly still a work in progress, entirely just polishing up the startup sequence a little bit. There will probably need to be checking to make sure that nothing gets regressed or broken, and there's CI errors that obviously need to get dealt with

guake/guake_app.py
Davidy22: I trimmed the condition because if you check new_page(), we can see that every page has a child so the condition is always true

Davidy22:
Alright, ironed out the silly mistakes that also failed CI, now just to put this into testing/checking

guake/guake_app.py
gsemet: Looks ok for me, do you want to change anything or we can merge it ?

Davidy22: I fixed what I found, PR stands for other people to find errors

Davidy22: If you've checked the code and tested it and can't find any issues, you can approve so this merges. We don't have to leave every PR in the queue for days

gsemet (approved):

PR #2012, opened 2022-01-18
Generate release notes for 3.8.2

Davidy22:
Also replaced a deprecated pandoc flag

Once this is merged, I'll tag the release. This has the upload to pypi on tag GA trigger, so hopefully this will finally update the pypi package. If there's any other changes that we want to add before we put the bow on this and push it out, I can regen the release notes after merging them. Off the top of my head, there might be some translation changes banked up in weblate that @gsemet might want to get in.

PR #2024, opened 2022-01-29
Generate release notes for 3.8.2

Davidy22:
Hopefully the proper release this time, unless someone decides to tag rc without saying anything again for no reason

Generated with `make release-note`, which was presumably how all the other release notes were generated. Note the lack of mysteriously deleted sections, which is still the main issue with #2015.

Will tag 3.8.2 when this is merged. If you merge this and tag a release, please don't leave the description empty again.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -8,9 +8,45 @@ guake
 Release Summary
 ---------------
 
+Eliminated redundant terminal spawning on startup
+
+Security Issues
+---------------
+
+- Fixed security Issue: Exposure of sensitive function, malicious user can arbitrary command via an execute_command d-bus method. #1796 
+
+Translation Updates
+-------------------
+
+- Chinese Simplified (@kitty-panics)
+
+- Croatian (@milotype)
+
+- Dutch (@Vistaus)
+
+- German (@m1ga, @rMb93)
+
+- Indonesian (@rezaalmanda)
+
+- Polish (@piotrdrag)
+
+- Russian (@vantu5z)
+
+- Swedish (@MorganAntonsson)
+
+- Turkish (@Draconis-25, @ersen0)
+
+3.8.2.0rc1
+==========
+
+Release Summary
+---------------
+
 Fix system font application issue
 
 Add fallback for version number finding
+
+Reinstated double clicking to open new tabs
 
 Deprecated pbr
 
@@ -20,6 +56,9 @@ New Features
 ------------
 
 - --is-visible option returns 1 when visible, and 0 when not
+
+- - Double click to open a new tab, without side effects in mouse
+    enabled terminal apps
 
 Known Issues
 ------------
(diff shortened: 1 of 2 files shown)


Davidy22:
@gsemet, @mlouielu, this doesn't have any code, is there anything holding up approving this?

mlouielu:
> 

Happy Lunar New Year 😸

Give me a second for this.

NEWS.rst
```diff
@@ -21,6 +57,9 @@ New Features
 
 - --is-visible option returns 1 when visible, and 0 when not
 
+- - Double click to open a new tab, without side effects in mouse
```
mlouielu, line 60: Is the duplicate `-` cause by original git commit?

mlouielu (changes requested):
LGTM, only one spot need to clarify

NEWS.rst
gsemet: It is just the slug file to edit (this can be done in a new commit)

Davidy22: The file is unmodified `make release-note` output. The dash comes from the reno file also having a dash.

Davidy22: File is restore_double_click_open_tab-c4dfc824ce4b5e85.yaml

mlouielu (approved):
LGTM

NEWS.rst
```diff
@@ -8,10 +8,46 @@ guake
...
+
+- Polish (@piotrdrag)
+
+- Russian (@vantu5z)
+
+- Swedish (@MorganAntonsson)
+
+- Turkish (@Draconis-25, @ersen0)
+
+3.8.2.0rc1
```
gsemet, line 39: even if the curernt release notes generator does not it, can we merge what it in the RCs and the current release?

Davidy22: We can do that by either not tagging RCs, removing the RC tag which may break some history, or patching the news file every time we generate it. Alternately, if you don't like seeing RCs in the news log, you can [not tag them on pulls that were intended to be releases](https://github.com/Guake/guake/pull/2012)

Davidy22:
Will tag 3.8.2 now that release notes are in

Davidy22:
Something about the pypi upload was messed up, might be pypi handling RCs badly. Next release we'll tag 3.9 without anyone jumping in and marking an RC and see if it works, if it marks as prerelease again then we might have a bad default set on the pypi key.

Also, how's the progress on readthedocs?

NEWS.rst
gsemet: I have what I describe at work: RCs are Ignored in the diff. So yes they can be removed, but they help test changes in the release process.

gsemet:
Looks like an issue with pbr or the checked out code in the pipeline, it has computed a wrong version in the GitHub action and uploaded a weird value. This is due to our inexperience with the GitHub workflow, not with pipy.

PR #2038, opened 2022-02-06
Force release version on tag pipeline

### LINKED ISSUE #2037: setuptools_scm still creaate .dev0 package
Because there are uncommited files during builds, setuptools_scm computes a .dev0 version ("dirty").

gsemet:
Fix #2037


gsemet:
@eli-schwartz can you check if this looks ok

Issue #928, opened 2017-04-03
On fullscreen mode don't appear on mouse display


pabloab:
When Guake is on fullscreen mode ignores the "Appear on mouse display" preference.

How to reproduce:
- Have at least two display (for eg a notebook to an external monitor through HDMI).
- Start Guake:
- Change your mouse cursor alternatively between displays and press F12 repeatedly
  - If it is in fullscreen mode: Will stay on the previous display.
  - If it is in window mode: Will appear on mouse display.

If this is the expected behavior _Appear on mouse display_ should warn about it and probably should have another preference checkbox to _Appear on mouse display **always**_.

I think this bug is not a duplicate of #607 where [@versusvoid](https://github.com/Guake/guake/issues/607#issuecomment-223942342) and [@alliefitter](https://github.com/Guake/guake/issues/607#issuecomment-283383585) mention this behavior.

VasuLief:
Having the same behaviour, but I actually like that and thought that this was a feature.

pabloab:
Don't you think should be clear on preferences _Appear on mouse display_ won't work if fullscreen? Also maybe someone (like me :-) ) want to have the possibility to use fullscreen on mouse display.
Perhaps that's the usual case when start using the _Start fullscreen_ preference:
![image](https://cloud.githubusercontent.com/assets/657836/24618512/aad4c66e-186d-11e7-85c6-eb8eb9783ae2.png)

langpavel:
I have trouble with fullscreen so I disabled it, but I also have problem with Appear on mouse display when I open guake on my (primary) display.
![image](https://cloud.githubusercontent.com/assets/142462/24703971/ec7d2630-1a05-11e7-89a4-fdd5de973bcb.png)

pabloab:
Maybe could be an option on context menu:
![image](https://cloud.githubusercontent.com/assets/657836/25190549/e971cfba-2502-11e7-8a40-65429efdb011.png)

austintackaberry:
I am also having this issue

PR #2049, opened 2022-03-01
fix fullscreen using pointer on several monitors

panpann:
This PR fixes the problem described in https://github.com/Guake/guake/issues/928 (at least on my computer)

I'm using several monitors with **dock bar** on all monitors and **top bar** on the main monitor.

It seems that when the guake's window height is larger than the height of the monitor excluding top/dock bars height, the monitor is sometimes "ignored" by gdk. If I'm resizing the guake window to be smaller than this height, it works.

The bug is at [`screen.get_monitor_at_point(x, y)`](https://github.com/Guake/guake/blob/master/guake/utils.py#L301).

It returns the number of the wrong monitor when I'm encountering the problem.

In this PR, I've reworked the method `get_final_window_monitor` to use latest gdk APIs (as pointed out by the TODO in the code) and instead of using [`.get_geometry()`](https://lazka.github.io/pgi-docs/Gdk-3.0/classes/Monitor.html#Gdk.Monitor.get_geometry), I'm using method [`.get_workarea()`](https://lazka.github.io/pgi-docs/Gdk-3.0/classes/Monitor.html#Gdk.Monitor.get_workarea) which seems to return the rectangle "usable" by an application (without dock / top bars or else)

I've also tested manually all guake's options of "Appear on mouse display" on single/multiple monitors and it works as expected.

I'm not used to guake and gdk libs so I don't know if it can break things or if you have specific "integration" tests to do other than tests in the lib ?

### DIFF
--- a/guake/utils.py
+++ b/guake/utils.py
@@ -39,7 +39,6 @@ from guake.globals import ALIGN_CENTER
 from guake.globals import ALIGN_LEFT
 from guake.globals import ALIGN_RIGHT
 from guake.globals import ALIGN_TOP
-from guake.globals import ALWAYS_ON_PRIMARY
 
 try:
     from gi.repository import GdkX11
@@ -230,9 +229,8 @@ class RectCalculator:
         log.debug("  vdisplacement = %s", vdisplacement)
 
         # get the rectangle just from the destination monitor
-        screen = window.get_screen()
         monitor = cls.get_final_window_monitor(settings, window)
-        window_rect = screen.get_monitor_geometry(monitor)
+        window_rect = monitor.get_workarea()
         log.debug("Current monitor geometry")
         log.debug("  window_rect.x: %s", window_rect.x)
         log.debug("  window_rect.y: %s", window_rect.y)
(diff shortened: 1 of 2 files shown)


guake/utils.py
```diff
@@ -285,34 +283,29 @@ def set_final_window_rect(cls, settings, window):
...
-        # If Guake is configured to use a screen that is not currently attached,
-        # default to 'primary display' option.
-        n_screens = screen.get_n_monitors()
-        if dest_screen > n_screens - 1:
-            settings.general.set_boolean("mouse-display", False)
-            settings.general.set_int("display-n", dest_screen)
-            dest_screen = screen.get_primary_monitor()
-
-        # Use primary display if configured
-        if dest_screen == ALWAYS_ON_PRIMARY:
```
Davidy22, line 312: I believe the always on primary option in the "Appear on display" drop-down is for using the system configured primary display if chosen, while other options in the drop-down let the user override with a specific monitor. Looks like the explicit check is removed here in favor of a catch-all if the search for a specific monitor falls through. Not noticing adverse side effects from the removal of this, raising this in case any potential edge case comes to mind.

Davidy22 (commented):
We do have tests for some things, this specific case is not one of the things covered by the test suite among other multi-monitor things. Might be nice to have. For now, the code looks alright and works for me as well, just want to check one thing before merging.

guake/utils.py
Davidy22: Looked fine, just wanted to make sure we didn't accidentally kill any of the defined monitor options.

Davidy22 (approved):

PR #2055, opened 2022-03-23
Save layout when splitting/closing split panes

### LINKED ISSUE #1686: A closed subterminal returns back after restarting Guake
When a terminal is splitted and then one of the subterminals is closed, it returns back after restarting Guake.
So it is virtually impossible to close a subterminal forever).
It only helps to close the tab.

Davidy22:
Resolves #1686


Davidy22:
That's a slightly annoying issue deepsource is raising, probably should be addressed in another issue though

Issue #2042, opened 2022-02-20
Security fix breaks `guake -e`


midnight-wonderer:
In version 3.8.5, `execute_command` has been removed from DBus
https://github.com/Guake/guake/issues/1796

The discussion there implies that `-e` option won't affect by this change.
In reality, it does break the functionality

Here the stack trace
~~~
Traceback (most recent call last):
  File "$HOME/.local/bin/guake", line 8, in <module>
    sys.exit(exec_main())
  File "$HOME/.local/lib/python3.10/site-packages/guake/main.py", line 633, in exec_main
    if not main():
  File "$HOME/.local/lib/python3.10/site-packages/guake/main.py", line 552, in main
    remote_object.execute_command(options.command)
  File "$HOME/.local/lib/python3.10/site-packages/dbus/proxies.py", line 72, in __call__
    return self._proxy_method(*args, **keywords)
  File "$HOME/.local/lib/python3.10/site-packages/dbus/proxies.py", line 141, in __call__
    return self._connection.call_blocking(self._named_service,
  File "$HOME/.local/lib/python3.10/site-packages/dbus/connection.py", line 652, in call_blocking
    reply_message = self.send_message_with_reply_and_block(
dbus.exceptions.DBusException: org.freedesktop.DBus.Error.UnknownMethod: Traceback (most recent call last):
  File "$HOME/.local/lib/python3.10/site-packages/dbus/service.py", line 662, in _message_cb
    (candidate_method, parent_method) = _method_lookup(self, method_name, interface_name)
  File "$HOME/.local/lib/python3.10/site-packages/dbus/service.py", line 254, in _method_lookup
    raise UnknownMethodException('%s is not a valid method' % method_name)
dbus.exceptions.UnknownMethodException: org.freedesktop.DBus.Error.UnknownMethod: Unknown method: execute_command is not a valid method
~~~

To reproduce, simply try `-e` option.

$guake --support
~~~
<details><summary>$ guake --support</summary>

Guake Version:		3.8.5

Vte Version:		0.60.3

Vte Runtime Version:	0.60.3

--------------------------------------------------
GTK+ Version:		3.24.20

GDK Backend:		<GdkX11.X11Display

--------------------------------------------------
Desktop Session: gnome

--------------------------------------------------
Display: :0

RGBA visual: True

Composited: True

* Monitor: 0 - XWAYLAND0
    * Geometry:		1920 x 1080 at 0, 0
    * Size:		480 x 270 mm²
    * Primary:		True
[...]

PR #2057, opened 2022-03-26
Reintroduce execute_command dbus hook

Davidy22:
Command now always creates a new tab to execute the command in

Alternate resolution to #2042 decided upon after some amount of discussion.

### DIFF
--- a/guake/dbusiface.py
+++ b/guake/dbusiface.py
@@ -153,6 +153,11 @@ class DbusManager(dbus.service.Object):
         self.guake.reset_terminal_custom_colors(current_terminal=True)
         self.guake.set_colors_from_settings_on_page(current_terminal_only=True)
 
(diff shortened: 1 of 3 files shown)


Davidy22:
@mlouielu I'm willing to use admin power to force my other PRs through but this one particularly I do want to get looked at due to security implications

Davidy22:
@mlouielu I've gone and forced the other PRs in limbo that aren't quite as sensitive as this one, still would like to get this one looked at before merge.

Davidy22:
@mlouielu I'm going to just push this if it comes close to a second month in limbo

Davidy22:
Just going to push this now. This spending as long as it did in limbo I think means I don't need to wait for review to merge things since anyone else who could review doesn't seem to be active

Issue #2060, opened 2022-03-28
Open links under terminal cursor


thibthib18:
For a more keyboard driven workflow, it would be great to be able to open links/URLs under terminal cursor via a keyboard shortcut.

I'd also be interested to make a PR if I could get some pointers. 

That functionality is already there for the mouse cursor, so I imagine there's a way to do this without too much effort if we can get the terminal cursor position.

Davidy22:
Keybindings get set  in keybindings.py, and documentation for the methods to interact with the VTE can be found [here](https://lazka.github.io/pgi-docs/Vte-2.91/classes/Terminal.html) if you were looking to put togethor a PR

Davidy22:
PR merged, closing

PR #2062, opened 2022-03-30
Add open link under terminal cursor

thibthib18:
Add functionality to set a keyboard shortcut to open link under terminal cursor (similar to "ctrl click").

Following [2060](https://github.com/Guake/guake/issues/2060)

Status: current changes are enough to open links under cursor via a hardcoded shortcut.

Question: how to properly add a new user configurable shortcut?

### DIFF
--- a/guake/keybindings.py
+++ b/guake/keybindings.py
@@ -82,6 +82,7 @@ class Keybindings:
             ("decrease-transparency", self.guake.accel_decrease_transparency),
             ("toggle-transparency", self.guake.accel_toggle_transparency),
             ("search-on-web", self.guake.search_on_web),
+            ("open-link-under-terminal-cursor", self.guake.open_link_under_terminal_cursor),
             ("move-tab-left", self.guake.accel_move_tab_left),
             ("move-tab-right", self.guake.accel_move_tab_right),
             ("switch-tab1", self.guake.gen_accel_switch_tabN(0)),
--- a/guake/terminal.py
+++ b/guake/terminal.py
@@ -478,6 +478,13 @@ class GuakeTerminal(Vte.Terminal):
         if value:
             return value
 
+    def get_link_under_terminal_cursor(self) -> Optional[str]:
+        cursor_position = self.get_cursor_position()
+        matched_string = self.match_check(cursor_position.column, cursor_position.row)
+        link = self.handleTerminalMatch(matched_string)
+        if link:
+            return link
+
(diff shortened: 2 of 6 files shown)


Davidy22:
You can add it to keybindings.py, stick it in the keys dict and test to see if the keybinding works and I'll check it after

guake/keybindings.py
```diff
@@ -82,6 +82,7 @@ def x(*args):
             ("decrease-transparency", self.guake.accel_decrease_transparency),
             ("toggle-transparency", self.guake.accel_toggle_transparency),
             ("search-on-web", self.guake.search_on_web),
+            ("open-link-under-terminal-cursor", self.guake.open_link_under_terminal_cursor),
```
Davidy22, line 85: A bit weird that this is in the middle like this but doesn't really matter I guess

guake/terminal.py
```diff
@@ -478,15 +478,23 @@ def handleTerminalMatch(self, matched_string):
         if value:
             return value
 
+    def get_link_under_terminal_cursor(self) -> Optional[str]:
```
Davidy22, line 481: Type hints would be a nice thing to do but probably should be consistent and everything else isn't hinted at the current moment, should probably be consistent with the rest of the file. Moving files onto hints could be a good bigger project though.

Davidy22 (changes requested):
Works for me and changes look mostly fine. Forgot to ask for a release note file, just do a `make reno SLUG=[descriptive_label]` and fill in the release note file like how it is in other PRs. Also a thing about the type hint

guake/terminal.py
thibthib18: Actually I wouldn't mind starting to add some more type hints to the codebase. Surely not everything, but a good chunk of the low hanging fruits just to get things started. If that's something you'd be interested in.

guake/keybindings.py
thibthib18: yeah agree doesn't matter much, just thought it fit better next to the other "Extras" category binding, than next to the more "terminal" related ones.

guake/terminal.py
Davidy22: Yeah that'd be nice, can take that if you're willing to do it. Just need the release note file generated by `make reno SLUG=[descriptive_label]` filled in and I'll merge

Davidy22 (approved):

Davidy22:
Alright looks fine, merging

PR #2063, opened 2022-03-31
Add TERM_PROGRAM environment variable like other well-known terminals

kohnish:
for shell run commands


Davidy22 (commented):
We can ignore the deepsource complaint, change looks fine, just need a release note file filled in, just do a `make reno SLUG=[descriptive_label]` and fill in the release note file like how it is in other PRs.

Davidy22 (approved):
Going to need to deal with the thing deepsource is complaining about in a bit, think it should be straightforward but that's another issue, I'll merge this.

PR #2076, opened 2022-04-20
Add file loader and tests for compute_tab_table and load_cwd_guake_yml

jorgeecardona:
This is a continuation of #1759. Test for compute_tab_title and load_cwd_guake_yml are added and a class that controls the cache is added to utils.py (FileManager) tests for that class are also included. There are no new functionalities here, just more tests and a cleaner cache control. Also dataclasses are not used in the code that's why I removed ir from the Pipfile.

### DIFF
--- a/guake/guake_app.py
+++ b/guake/guake_app.py
@@ -19,7 +19,6 @@ Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
 Boston, MA 02110-1301 USA
 """
 import json
-import yaml
 import logging
 import os
 import shutil
@@ -70,6 +69,7 @@ from guake.simplegladeapp import SimpleGladeApp
 from guake.theme import patch_gtk_theme
 from guake.theme import select_gtk_theme
 from guake.utils import BackgroundImageManager
+from guake.utils import FileManager
 from guake.utils import FullscreenManager
 from guake.utils import HidePrevention
 from guake.utils import RectCalculator
@@ -191,9 +191,8 @@ class Guake(SimpleGladeApp):
         # FullscreenManager
         self.fullscreen_manager = FullscreenManager(self.settings, self.window, self)
 
-        # Hold a copy of guake_yaml
-        self._guake_yml = {}
-        self._guake_yml_load_monotonic = {}
+        # Start the file manager (only used by guake.yml so far).
+        self.fm = FileManager()
 
         # Workspace tracking
         self.notebook_manager = NotebookManager(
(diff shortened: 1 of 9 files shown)


guake/guake_app.py
```diff
@@ -1113,23 +1113,25 @@ def load_cwd_guake_yaml(self, vte) -> dict:
...
-                    with guake_yml.open(encoding="utf-8") as fd:
-                        content = yaml.safe_load(fd)
-                        self._guake_yml[guake_yml] = content
-                        self._guake_yml_load_monotonic[guake_yml] = pytime.monotonic()
-            except PermissionError:
-                log.debug("PermissionError on accessing .guake.yml")
+        content = None
+        try:
+            content = self.fm.read_yaml(filename)
+        except PermissionError:
```
Davidy22, line 1121: These look like exceptions that can always occur with read_yaml(). Maybe these should be inside the function, so that future potential calls don't need to also wrap it in this chunk of exception handling?

Davidy22 (changes requested):
I notice the last three commits in this PR are fixing the tests from the commit before them, and there's some duplicate names too. Before we merge, the last four commits can probably be squashed down so we just have one commit adding the tests in main. Besides that, the other commits seem like fine subdivisions.

guake/guake_app.py
jorgeecardona: Sure, there are pleny of possible errors when reading files and expecting the content to have certain format. I just can't think of a good value to return when there is an error with the format, and it seems to me that returning an empty dict would obscure the fact that the file is incorrectly formatted. I could raise a ValueError I guess.

Unless you have a suggestion for a default value to return I would leave it as it is.

Davidy22: A None, or a single encapsulating exception, both fine

jorgeecardona: The method `read_yaml` is now returning None in the case of any "expected" error, like PermissionError , NotFoundFile, Encoding or YAMLError. If anything else happens the exception would be capture by the code in `guake_app.py`

Davidy22 (approved):
Looks good, we can get away with no release notes on this one because it's pretty much part of the earlier PR. There was a strange issue a bit ago that was making CI fail that seems to be a crash bug that didn't appear until after I started merging unrelated PRs, but I went and dealt with that in 90b9bfad2057ce66e60e23af13b1c10ff2de1375

Davidy22:
Actually, I've been thinking about being a bit more liberal with review privileges than the project's been so far. Current repo setting is PRs need an approving review and I'm ok with that, but reviews happen quite infrequently when it's not me and I kind of want to get more people in here. Since you have a few PRs now if you want, I can give you review privileges, and it'll be up to you how often you check in and look at something on the PR queue.

jorgeecardona:
Great, I noticed your change in #90b9bfa and moved the initialization of the FileManager at the same point. I didn't notice any probem before, but I guess the current tests and how I manually tested didn't capture some cases where a title is loaded before that in `__init__`.

PR #2101, opened 2022-06-18
Close select tab by label / select tab by label

Arwez:


Davidy22:
Looks like you need to do a passthrough of black to make CI pass, and maybe some more for tests after that. CI should run on your repo too, so you can see the results of most of the runs without needing to wait for me to approve the run over here.


As for technical changes, this PR has made me aware that while we have a close_tab defined and written, it's not actually connected to a front facing guake flag for whatever reason. If we presume that oversight didn't exist and we did have a close tab flag, I believe the close_tab_by_label() in guake_app is unnecessary because we can functionally duplicate it in the dbusiface by calling select_tab_by_labe()l followed by a close_tab(). I can add the flag for a plain --close-tab flag myself or you could fold that into here as a thematically similar function if you want to.

ianpilogov:
Any updates on this feature? I'd like to close tabs from CLI interface for my automation, but looks like guake can't do it.

Davidy22:
Somewhat busy personally but if you're willing to fork and fix up this branch and PR it I can merge it if it looks good and passes CI

novaTopFlex (approved):
Agree with the code; the code looks appropriate to the situation and correcrtions have been made successfully!

Issue #2088, opened 2022-05-12
feature: ability to use custom search engine for "Search on Web" (hardcoded to Google)


ElijahLynn:
The "Search on Web" feature is very useful. I recently switched to a non-Google search engine and noticed that Guake is hardcoded to use Google. It would be nice to have this be user customizable. 

https://github.com/Guake/guake/blob/2cfc9b3ab00f8288fb119c1c6f663552384732cf/guake/callbacks.py#L50-L58

https://github.com/Guake/guake/blob/0db81488d17e34df52c45e75b51d022a118dd9d3/guake/guake_app.py#L1323-L1336

PhantomLel:
can i take this?

oewami:
yes.

Davidy22:
If you want to work on resolving an issue, you can just send in a PR and if it works and looks good I'll merge.

hackedd:
Added with #2134.

PR #2125, opened 2022-09-07
search engine can be set to preference

PhantomLel:
For issue #2088


PhantomLel:
why's that failing?

Davidy22:
It should be possible to run actions on your own repo so that you don't need to wait for me to come round and hit the approve button on this for CI, I'll look back some time later when I remember to to see the result of this next run

PhantomLel:
Ok thanks looks like I just had to run `black`

guake/data/prefs.glade
```diff
@@ -135,7 +135,7 @@
                         <property name="can-focus">False</property>
                         <property name="halign">start</property>
                         <property name="margin-top">6</property>
-                        <property name="label" translatable="yes">&lt;span size="18000"&gt;&lt;b&gt;Guake properties&lt;/b&gt;&lt;/span&gt;</property>
```
Davidy22, line 138: Typo introduced?

guake/data/prefs.glade
```diff
@@ -689,6 +698,49 @@
...
+                            <child>
+                              <object class="GtkBox">
+                                <property name="visible">True</property>
+                                <property name="can-focus">False</property>
+                                <property name="spacing">4</property>
+                                <child>
+                                  <object class="GtkLabel" id="lb_search_engine">
+                                    <property name="visible">True</property>
+                                    <property name="can-focus">False</property>
+                                    <property name="label" translatable="yes">Search Engine:                         </property>
```
Davidy22, line 710: Generally don't want to do spacing by just adding a bunch of spaces when doing UI work. I assume you did this to make a UI element line up somewhere, but if you want to do that you should use columns. The spaces are difficult for the translators to understand and match, just strip the spaces and if you want to do a re-layout that can be another pr

Davidy22:
Added a release note file, the patch works, for the future it could do with a way for the user to add their own search engine but this'll do for now I think

Davidy22 (approved):

PR #2133, opened 2022-10-14
Add Neeva.com search engine

ElijahLynn:
Follow up to https://github.com/Guake/guake/issues/2088 by @PhantomLel.

### DIFF
--- a/guake/callbacks.py
+++ b/guake/callbacks.py
@@ -17,6 +17,7 @@ ENGINES = {
     1: "duckduckgo.com/",
     2: "bing.com/search?q=",
     3: "yandex.com/search?text=",
+    4: "neeva.com/search?q=",
 }
 
 


Davidy22 (changes requested):
I do want to eventually shift towards the thing I mentioned in the original pull request where we have a way to enter your own search engine in preferences because I really don't want the endgame for this to be a mile long list of search engines, but I won't say no to additions. This PR does need a little revision though, because as it is currently, the search engine won't actually show up in the list of options in preferences. Need to mirror the changes in [the prefs.glade file](https://github.com/Guake/guake/pull/2125/files#diff-898b51cf548cdc0e879567351059f66ace331de76143a51007e8d140539da766), put your addition in the same relative spot in the drop down list so that people can actually choose it. You should be able to install and test your changes with a `sudo make install`

PhantomLel:
I can work on allowing users to add whatever search engine

Davidy22:
Oh certainly that would be very welcome

Davidy22:
Closing because superceded by #2134

ElijahLynn:
W00t! Yeah, much better to go with user entered, custom search engines! Thank you!

PR #2134, opened 2022-10-18
Added custom search engine input #2088

PhantomLel:
Added input on the general menu to add a custom search engine.

### DIFF
--- a/guake/callbacks.py
+++ b/guake/callbacks.py
@@ -9,15 +9,8 @@ from guake.prefs import PrefsDialog
 from guake.utils import FullscreenManager
 from guake.utils import HidePrevention
 from guake.utils import get_server_time
+from guake.globals import ENGINES
 from urllib.parse import quote_plus
-
-# the urls of the search engine options
-ENGINES = {
-    0: "www.google.com/search?safe=off&q=",
-    1: "www.duckduckgo.com/",
-    2: "www.bing.com/search?q=",
-    3: "www.yandex.com/search?text=",
-}
 
 
 class TerminalContextMenuCallbacks:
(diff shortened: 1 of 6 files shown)


guake/callbacks.py
```diff
@@ -62,12 +62,19 @@ def on_search_on_web(self, *args):
...
-                    "https://www." + ENGINES[self.settings.general.get_int("search-engine")] + query
-                )
-                Gtk.show_uri(self.window.get_screen(), search_url, get_server_time(self.window))
+            # nothing selected
+            if not query:
+                return
+
+            selected = self.settings.general.get_int("search-engine")
+            # if custom search is selected, get the engine from the 'custom-search-engine' setting
+            if selected == 4:
```
Davidy22, line 71: The check for the custom option should be checking to see if it's the last index. Hardcoding to 4 can generate unexpected bugs if the list of preset search engines changes length in the future.

guake/prefs.py
```diff
@@ -1050,7 +1065,15 @@ def load_configs(self):
 
         # search engine
         value = self.settings.general.get_int("search-engine")
+        custom_search = self.get_widget("custom_search")
+        custom_search.set_text(self.settings.general.get_string("custom-search-engine"))
         self.get_widget("search_engine_select").set_active(value)
+        # if 'Custom' is selected make the search engine input editable
+        if value == 4:
```
Davidy22, line 1072: Similar deal here

Davidy22 (changes requested):
Works great and mostly looks good, just one spot that could probably be adjusted for safety

PhantomLel:
I can't find anyway to get the length of the ComboBox. Do you know of a method to do that?

Davidy22:
You can query the length of a combobox if you already have a pointer to it, but to my knowledge it's a bit ugly getting that when the combobox is nestled deep in a glade layout because there's no get element by id equivalent, but we also have the `ENGINES` dict that you defined that should mirror the length of the available options pretty reliably that you can use.

PhantomLel:
Alright that's what I was thinking. Thanks!

Davidy22 (approved):
Looks good, merging. Thanks for the contributions!

PR #2137, opened 2022-10-24
Makefile venv cosmetic fixes

igoropaniuk:

- rebase on latest HEAD:

  ```bash
  ```

- hack your change

  ```bash
  ```

  generation, using:

  ```bash
  ```

- create new commit message

  ```bash
  $ <hack the code>
  $ git commit --all
  ```

  line in your commit messages:
  
  ```
  # For a bug fix, uses:
  sem-ver: bugfix
  
  # For a new feature, uses:
  sem-ver: feature
  
  ```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ ln-venv:
 	ln -s $$(pipenv --venv) .venv
 
 clean-ln-venv:
-	@rm -f .venv
+	@rm -rf .venv
 
 install-system: install-schemas compile-shemas install-locale install-guake
 
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -18,6 +18,7 @@ colorlog
 fiximports>=0.1.18
 flake8>=3.8.0,<4.0.0
 flakehell
+dbus-python
 mock>=2.0.0
 pathlib2
 pep8
@@ -25,10 +26,12 @@ pew
 pipenv-to-requirements
 pyfakefs
 pyflakes
+PyGObject
 pylint
 pytest
 pytest-cov
 pytest-mock
+pyyaml
 reno[sphinx]>=2.8.0
 scandir ; python_version > '3.5'
 sphinx-rtd-theme


Davidy22 (changes requested):
The stuff you added to the requirements file I believe need to come from non-python distro packages, and it looks like there's some dependency issues blocking them in the test suite. The venv change seems fine, never actually used that makefile option but seems like it wouldn't have actually worked in its previous state

igoropaniuk:
@Davidy22 The problem is that CI runner doesn't install packages listed in `bootstrap-dev-<dist>.sh` ( such as `python-dbus`, `python3-gi` etc), nor installs these packages via `pip` into venv (at least what I see in `.github/workflows/ci.yml`)

My initial assumption was that I could use `requirements-dev.txt` for local development/testing (based on `-dev` prefix), but looks like it's used only by CI job, which just runs tests in `guake/tests` dir (just correct me, if I'm wrong; I personally prefer to use simple venv instead of Pipenv).


Won't you mind if I address that issue by addiitionall adding install steps for `libdbus-1-dev libdbus-glib-1-dbus` in `.github/workflows/ci.yml` for 


As there are bunch of places where `PyGObject` is imported:
```
guake.git$ grep -e "import gi" -r ./guake/
./guake/guake_app.py:import gi
./guake/split_utils.py:import gi
./guake/main.py:        import gi
./guake/main.py:        import gi
./guake/__init__.py:    import gi
./guake/__init__.py:    import gi
./guake/__init__.py:    import gi
./guake/prefs.py:import gi
./guake/utils.py:import gi
./guake/settings.py:import gi
./guake/simplegladeapp.py:import gi
./guake/keybindings.py:import gi
./guake/terminal.py:import gi
./guake/support.py:import gi
./guake/notifier.py:import gi
./guake/dialogs.py:import gi
./guake/gsettings.py:import gi
./guake/theme.py:import gi
./guake/notebook.py:import gi
./guake/about.py:import gi
./guake/callbacks.py:import gi
./guake/boxes.py:import gi
./guake/menus.py:import gi
./guake/customcommands.py:import gi
./guake/common.py:import gi

```


BTW, what the reason to use two different requirements files (regular and `-dev`)?

Davidy22:
Seperate -dev file is there so that there's a clean requirements file that only contains the dependencies needed to actually run the program for people who don't want to do any of the dev stuff. This line of questioning also reminded me to check and remember that the two requirements files are auto-generated, and shouldn't be manually edited like this PR is doing.

Steps to carry out the non-pypi installation steps probably wouldn't hurt, general health check and all. Currently no tests rely on them, but it opens the door for expansion on that front I think.

PR #2138, opened 2022-10-25
Add pin button to temporarily prevent hiding when losing focus

### LINKED ISSUE #2089: feature / pin button - temporary keep focus
It could be great to have "pin" button on bottom right side, near "new tab" button. Like on Yakuake 
image

Also some smart close on lost focus function could be implemented. 
For example its really easy make screenshot with Yakuake. Guake with enabled losing focus cannot be screenshoted.

hackedd:
Closes #2089.


robinpecha:
Im using latest 3.9.0 but I cannot find how to show this button anywher, please do you have some hint? Thank you

Issue #2156, opened 2023-01-25
Unable to run `make build` with pipenv 2022.12.19


theCapypara:
**Background:**

I wanted to add something to Guake and thus wanted to checkout the code locally and set it up for development. For that I had a look at the CI and tried to run all make commands used there to make sure I am having a stable environment to work with.

**Describe the bug**

I am however unable to run `make build` locally without issues.

**Expected behavior**

When checking out the project from source, as described in the docs ( https://guake.readthedocs.io/en/stable/user/installing.html#install-from-source )I expect `make build` to run through without errors after installing dependencies via `pipenv install --dev`.

**Actual behavior**

The `pipenv_to_requirements` step fails:

```
PIPENV_IGNORE_VIRTUALENVS=1 pipenv run pipenv_to_requirements
Traceback (most recent call last):
  File "/home/marco/.local/share/virtualenvs/guake-bcq_gztv/bin/pipenv_to_requirements", line 8, in <module>
    sys.exit(main())
  File "/home/marco/.local/share/virtualenvs/guake-bcq_gztv/lib/python3.10/site-packages/pipenv_to_requirements/__init__.py", line 115, in main
    def_req = parse_pip_file(pipfile, 'default')
  File "/home/marco/.local/share/virtualenvs/guake-bcq_gztv/lib/python3.10/site-packages/pipenv_to_requirements/__init__.py", line 66, in parse_pip_file
    return [clean_version(n, i) for n, i in pipfile.get(section, {}).items()]
AttributeError: 'function' object has no attribute 'get'
```

Upstream issue: https://github.com/gsemet/pipenv-to-requirements/issues/25 
Upstream PR: https://github.com/gsemet/pipenv-to-requirements/pull/26

**To Reproduce**

See above + see below for my installed package versions.

OS: Arch Linux
Python: 3.10.9
pipenv: `2022.12.19` (The last successful build on Github Actions used `2022.11.30` for reference - referring back to the issue above, I'm surprised that this works. The issue might not simply be the pipenv version.)

Installed packages in the venv:
```
alabaster==0.7.13
astroid==2.13.3
attrs==22.2.0
autopep8==1.5.7
Babel==2.11.0
black==21.8b0
bleach==6.0.0
certifi==2022.12.7
cffi==1.15.1
cfgv==3.3.1
charset-normalizer==3.0.1
click==8.1.3
colorama==0.4.6
colorlog==6.7.0
coverage==7.1.0
cryptography==39.0.0
[...]

Issue #1806, opened 2020-10-22
Guake does not display images when w3m-img is installed


fluffapparition:
Be sure to check the existing issues (both open and closed!), and make sure you are running the latest version of Guake.

For how to run the latest Guake in your computer, please refer to [Install from source](https://guake.readthedocs.io/en/latest/user/installing.html#install-from-source).

**Describe the bug**

Guake does not display images when w3m-img is installed

**Expected behavior**

Google Logo appears

**Actual behavior**

Text only

**To Reproduce**

Install w3m and w3m-img through your package manager, then run `w3m google.com` in both Guake and another terminal (Konsole in this case)

-------------------------------------------------------------------------------

Please run `$ guake --support`, and paste the results here. Don't put backticks (`` ` ``) around it! The output already contains Markdown formatting. And make sure you run the command **OUTSIDE** the Guake.

<details><summary>$ guake --support</summary>

Guake Version:          3.6.3

Vte Version:            0.60.3

Vte Runtime Version:    0.60.3

--------------------------------------------------
GTK+ Version:           3.24.20

GDK Backend:            GdkX11.X11Display

--------------------------------------------------
Desktop Session: plasma

--------------------------------------------------
Display: :0

RGBA visual: True

Composited: True

* Monitor: 0 - DEL HDMI-0
    * Geometry:         1920 x 1080 at 0, 0
    * Size:             509 x 286 mm²
    * Primary:          True
    * Refresh rate:     60.00 Hz
    * Subpixel layout:  unknown

Davidy22:
Did some cursory research, figured out that $WINDOWID needs to be set for this to work. Both gnome-terminal and guake had this issue, and finding window id with xwininfo and running that through a hex-dec converter then setting it to $WINDOWID fixed the issue completely on gnome-terminal, rendered weirdly and in the wrong place with guake though. Can look further into this, although I did notice [some other](https://github.com/Tetralet/LilyTerm/issues/61) maintainers of [other terminals](https://github.com/gnunn1/tilix/issues/540) talking about the implementation being a bit hacky, which inclines me to rather hope for w3m to adopt sixel rendering which we should inherit by default with VTE.

czg-sci-42ver:
I tried `ranger `by guake , however , I failed .
According to [this issue comment ](https://github.com/ranger/ranger/issues/2429#issuecomment-906765007)

### does guake support `Kitty's image protocol` or `SIXEL`?

- if so ,what should I config ? 

[...]

PR #2157, opened 2023-01-25
Add Sixel Support

theCapypara:
This adds support for displaying Sixels, in case VTE is compiled with support for it enabled.
This is related to #1806.

Upstream issue for Sixel support: https://gitlab.gnome.org/GNOME/vte/-/issues/253

***Side notes:***
While developing I came across three issues:
- #2156 
- I needed to pin pylint to `2.13` because otherwise it would complain about options that no longer exist.
- I needed to pin Click to `8.0.4`: https://github.com/psf/black/issues/2964


Davidy22:
Sorry about the delay, I've dropped some older unsupported python versions from CI so it should run if you push a superficial change and I'll take a look in the meantime

guake/terminal.py
```diff
@@ -211,6 +211,15 @@ def configure_terminal(self):
         if hasattr(self, "set_alternate_screen_scroll"):
             self.set_alternate_screen_scroll(True)
 
+        enable_sixel = client.get_boolean("use-sixel")
```
Davidy22, line 214: Does this need to be a preferences option? Seems like the kind of thing that anyone who has a terminal that can support it would want. If we do keep this as an option in preferences, we also need to add some switches on setting toggle, because as-written currently if a user clicks the box to enable they still won't see sixels until they restart guake.

Davidy22 (commented):
Been busy, but I got the time to take a look at this, don't have a sixel enabled version of VTE installed to check that the flag works so I still need to do that but I do still have thoughts on parts of this.

guake/terminal.py
theCapypara: I figured it's probably a good idea to have this optional at least as long as sixel support is still considered experimental in vte.

Davidy22: While it is experimental, it's also quite unavailable, I think the only way currently for people to have a version of general VTE installed is to compile it yourself, which is tantamount to opting in, and when it lands in an official release pushed to general users I would think it is no longer experimental.

theCapypara: That makes sense. I'll remove the option to enable them and just turn them on when available.

theCapypara: I'd probably still keep the setting in the gschema, if advanced users want to still opt-out somehow / for some reason. But I'll default it to enabled, would that be okay?

Issue #2205, opened 2023-10-07
Prevent users from setting non-image files as a background image


Davidy22:
As noticed in #2198, this is currently possible and causes a crash bug. Add a check to stop users from setting non-image files in the background selection dialog

HeetVekariya:
Is this issue solved or not ?
If not then can you please locate the file/functions which is to corrected such that it only allows the image file to be set as background files.

kapami:
Hello, I would like to work on this issue. Can you please assign me to this issue?
Thank you

Davidy22:
You don't need to wait to be assigned anything, just send in a PR that fixes an issue and link it.

HeetVekariya:
@Davidy22 how can i connect with you on the discord ?

Davidy22:
Is there something you want to discuss that isn't related to this PR?

HeetVekariya:
I want to know what is to be changed such that the issue is solved. For that i want to join the discord of this repo if exist.

Davidy22:
Anything relevant to this issue can be discussed here. You should be able to find what you need to change by searching for the word background, probably just need to change the code around the dialog in prefs.py

PR #2206, opened 2023-10-08
Prevent the use of non-image files as a background image

0scvr:
Please follow these steps before submitting a new Pull Request to Guake:

- rebase on latest HEAD:

  ```bash
  $ git pull --rebase upstream master
  ```

- hack your change

- to execute the code styling, checks and unit tests:

  ```bash
  $ make style check reno-lint test
  ```

- describe your change in a slug file for automatic release note
  generation, using:

  ```bash
  $ make reno SLUG=<short_name_of_my_feature>
  ```

  and edit the created file in `releasenotes/notes/`.
  You can see how `reno` works using `pipenv run reno --help`.

  Please use a generic slug (eg, for translation update,
  use `translation`, for bugfix use `bugfix`,...)

- create new commit message

  ```bash
  $ <hack the code>
  $ git commit --all
  ```

- If your change is related to a GitHub issue, you can add a reference
  using `#123` where 123 is the ID of the issue.
  You can use `closes #123` to have GitHub automatically close the issue
  when your contribution get merged

- Semantic commit is supported (and recommended). Add one of the following
  line in your commit messages:

  ```
  # For a bug fix, uses:
  sem-ver: bugfix

  # For a new feature, uses:
  sem-ver: feature

  # Please do not use the 'breaking change' syntax (`sem-ver: api-break`),
  # it is reserved for really big reworks
  ```

Fixes #2205

PR #2207, opened 2023-10-10
Prevent the use of non-image files as a background image (replaces #2206)

0scvr:
Fixes https://github.com/Guake/guake/issues/2205 and replaces PR #2206.

I had an issue with my setup so I thought it would cleaner to start over in a new branch and port my changes there.


Davidy22 (approved):
Looks fine, merging.

Issue #1931, opened 2021-10-15
Add a  --split-horizontal-height=XX  and  --split-vertical-width=XX


Lowrida:
A suggestion

Add a --split-horizontal-height=XX  (XX= lines)  for max height of new terminal
Add a --split-vertical-width=XX (XX= cols)  for max width of new terminal

I use a script to open a new tab that opens 2 horizontal splits (3 terminals).
First one is just 1 line and shows clock, cpu, mem etc info every second. 
Second one runs midnight commander in it with output to third terminal (exec 1>/dev/pts/xx).
Third one is the output of the second terminal.

But i have to drag the size of the terminals manually every time i open an new tab, because the first split is halve of the screen while terminal 1 output is just a one liner.

I couldn't find anything on the net to set this on the cli.

Screenshots:
![guake-split-a](https://user-images.githubusercontent.com/53632533/137553810-4ddb4583-097f-4d35-bba2-f4d14d2a3f9b.png)
.
![guake-split-b](https://user-images.githubusercontent.com/53632533/137553825-afd3b19f-e39a-46d9-905d-daed60c7abdc.png)

Lowrida:
I made a quick&dirty script to do what i want, but xdotool is slow as F#! :/  I still hope these options will be implemented in an upcoming release..

Maybe usable for anyone with the same problem:
`#!/bin/sh
######################################################################################################################
#                                                                                                                    #
#   This script opens a new tab, splits it in three, resizes each terminal and starts a program in each terminal.    #
#   The values of the sleep timers depends on the speed of your system. Untill there is an option to set the size    #
#   of the new split, this is a slow&dirty solution. Feel free to (ab)use it as you like!                            #
#                                                                                                                    #
#                                                                                                                    #
######################################################################################################################
#
# Open new tab and terminal
guake -new-tab=/home/admin -e 'echo "Terminal0: 0" >/tmp/guake_terminal.tmp; c0=$(tty); echo "Console0: $c0" >>/tmp/guake_terminal.tmp; lines0=$(tput lines); echo "Lines0: $lines0" >>/tmp/guake_terminal.tmp'
#
# Open second terminal find out number and tty
[...]

PR #2208, opened 2023-10-11
Allow splitting with specified size using --split-vertical or --split-horizontal

welhoilija:
Introduces the ability to specify split percentages when using the --split-vertical or --split-horizontal flags, as discussed in #1931. Now, users can control the size of the new pane by providing an optional split percentage, making terminal splitting more flexible and user-friendly.

### DIFF
--- a/guake/main.py
+++ b/guake/main.py
@@ -240,17 +240,29 @@ def main():
     parser.add_argument(
         "--split-vertical",
         dest="split_vertical",
-        action="store_true",
-        default=False,
-        help=_("Split the selected tab vertically."),
+        metavar="SPLIT_PERCENTAGE",
+        action="store",
+        type=int,
+        nargs="?",
+        const=50,
+        default=None,
+        choices=range(1, 100),
+        help=_("Split the selected tab vertically. Optional input split percentage for the width."),
     )
 
     parser.add_argument(
         "--split-horizontal",
         dest="split_horizontal",
-        action="store_true",
-        default=False,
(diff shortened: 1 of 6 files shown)


guake/main.py
```diff
@@ -240,17 +240,29 @@ def main():
...
-        action="store_true",
-        default=False,
-        help=_("Split the selected tab vertically."),
+        metavar="SPLIT_PERCENTAGE",
+        action="store",
+        type=int,
+        nargs="?",
+        const=50,
+        default=None,
+        choices=range(101),
```
Davidy22, line 249: The range needs to be constrained on the edges. Math error at choice = 0, and the behavior at 100 may be unintuitive, takes up the whole tab but still leaves the previous split in the tab hidden and crammed away. Trimming the range to 1-99 should mop these up, and maybe should be 5-95 but maybe there's someone out there with a giant screen who wants a 98% split, I guess 1-99 is fine.

Davidy22 (changes requested):
Works nicely, passes checks, but some edge cases to consider.

guake/main.py
welhoilija: Good catch! I changed it to range(1,100), which limits the options to 1-99 for both horizontal and vertical..

welhoilija:
Sry, will run the formatter once i get home, my bad.

welhoilija:
Should be good now, I think there was also some release note that had some trailing whitespace.

Davidy22 (approved):
Seems good now, merging.

Issue #1958, opened 2021-10-29
Save content saves only visible screen


filipjenicek:
**Describe the bug**
The "Save content..." option in the context menu saves only the visible content of the terminal.

**Expected behavior**
Full terminal output is saved including scrolled content.

**Actual behavior**
On Debian 11, only the visible part is saved.
AFAIK on Debian 10 all contents were saved correctly.

**To Reproduce**
Debian 11 (Oct29) xfce4
Reproduced with debian packages 3.4.0-1, 3.8.0-1 and also with the latest src code a084a1d0d11d2306ca5447c2cfaf31db0393c798.

<details><summary>$ guake --support</summary>

Guake Version:		3.8.2.dev4

Vte Version:		0.62.3

Vte Runtime Version:	0.62.3

--------------------------------------------------
GTK+ Version:		3.24.24

GDK Backend:		<GdkX11.X11Display

--------------------------------------------------
Desktop Session: lightdm-xsession

--------------------------------------------------
Display: :0.0

RGBA visual: True

Composited: True

* Monitor: 0 - DEL DVI-0
    * Geometry:		1920 x 1200 at 0, 0
    * Size:		518 x 324 mm²
    * Primary:		True
    * Refresh rate:	59.95 Hz
    * Subpixel layout:	horizontal-rgb

filipjenicek:
You can reproduce it on Debian 11:
1. Download https://www.linuxvmimages.com/images/debian-11/
2. #sudo apt install guake
3. open guake, execute sudo dmesg
4. Run the guake context menu action "save content..."
5. The resulting file has only a few lines depending on your screen size.

It works well on Debian 10:
https://www.linuxvmimages.com/images/debian-10

The key difference is the version of libvte, which is 0.62.3 on Debian 11, and 0.54.2-2 on Debian 10.

If you downgrade libvte to 0.54.2-2 (https://packages.debian.org/buster/libvte-2.91-0), it starts to work well. I suppose the guake's clipboard hack is not compatible with newer releases of libvte.

Davidy22:
Don't have a newer version of VTE in my repo, but since this is isolated to newer VTE I should be able to figure this out, can start checking documentation for now.

Davidy22:
Now have vte 0.66.2 and still not encountering this, although I am observing some kind of hang that's probably a different issue.

nyaapa:
Reproes on ubuntu 22.04 w/ both stock (3.8.5) and updated from linuxuprising (3.10) guake.
[...]

PR #2218, opened 2023-12-18
Fix save content with newer Vte

smith-doug:
Fixes Save Content feature with newer versions of Vte.  The entire contents, including scrollback, will be saved.

See #1958   I left the old method there as well, in case get_text_range doesn't work for some reason.  I don't have a system with an older version of Vte to test.

### DIFF
--- a/guake/dialogs.py
+++ b/guake/dialogs.py
@@ -143,14 +143,24 @@ class SaveTerminalDialog(Gtk.FileChooserDialog):
         self.parent_window = window
 
     def run(self):
-        self.terminal.select_all()
-        self.terminal.copy_clipboard()
-        self.terminal.unselect_all()
-        clipboard = Gtk.Clipboard.get_default(self.parent_window.get_display())
-        selection = clipboard.wait_for_text()
-        if not selection:
-            return
-        selection = selection.rstrip()
+        # row from get_cursor_position takes the scrollback rows into account automatically
+        row = self.terminal.get_cursor_position()[1]
+        # Get the max col, not where the cursor is, to avoid losing text at the end
+        col = self.terminal.get_column_count()
+        content = self.terminal.get_text_range(0, 0, row, col)
+
+        if content:
+            selection = content[0].rstrip().lstrip()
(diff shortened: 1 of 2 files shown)


Davidy22:
Sorry about the slow reply, been bogged down for a bit. Built and ran, and I get the following traceback:

```
  File "/usr/lib/python3.12/site-packages/guake/callbacks.py", line 39, in on_save_to_file
    SaveTerminalDialog(self.terminal, self.window).run()
  File "/usr/lib/python3.12/site-packages/guake/dialogs.py", line 153, in run
    selection = content[0].rstrip().lstrip()
                ^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'rstrip'
```

My terminal I presume is an older version of VTE because I'm able to use save content for the full terminal history without the patch. I suspect that the condition in the if statement reliant on get_text_range() may need to be reworked.

guake/dialogs.py
```diff
@@ -143,14 +143,24 @@ def __init__(self, terminal, window):
...
-        if not selection:
-            return
-        selection = selection.rstrip()
+        # row from get_cursor_position takes the scrollback rows into account automatically
+        row = self.terminal.get_cursor_position()[1]
+        # Get the max col, not where the cursor is, to avoid losing text at the end
+        col = self.terminal.get_column_count()
+        content = self.terminal.get_text_range(0, 0, row, col)
+
+        if content:
```
PhungXuanAnh, line 152: I add this  suggestion to fix this error: https://github.com/Guake/guake/pull/2218#issuecomment-2134558574
```suggestion
        if content and content[0]:
```

Davidy22: Looks like this does fix it.

Davidy22 (approved):
Patch looks like it works now, but there's some unfortunate behavior with the way this interacts with the hide on focus loss option. It seems like a common issue across everything that pops a dialog, maybe we need a utlity wrapper for dialogs that'll temporarily disable hide on lose focus. That'd be out of scope for this PR though, merging this.

Issue #2225, opened 2024-02-06
[Feature] Only underline quick open targets when quick-open is enabled.


joeperri95:
One of the regexes used to match quick open targets will also match certain expressions in rust, C++ and I imagine any other languages that make use of the double colon notation.

This can sometimes cause a significant portion of my screen to be underlined when using guake to view or edit code with a terminal application.

It would be nice if the underlining was disabled when the quick-open setting is not set to True.

PR #2235, opened 2024-03-26
Remove underlines from quick-open targets when quick-open disabled

chrhansk:
Fixes https://github.com/Guake/guake/issues/2225


joeperri95:
This seems to resolve the behaviour I was seeing when I opened #2225. Thanks!

Davidy22 (approved):
Seems mostly good. There's an extra spellcheck change that isn't strictly related to the other fix, usually good to split up unrelated things for a sort of atomicity for the sake of ease of reverting and history reading but eh it's just a little typo fix

PR #2236, opened 2024-03-27
added background color and text color to tab label

selam:
this patch adds custom background and text color to tabs,

i am not good  at gtk library and related tech, i just want to custom colored tabs, instead of label maybe for tab style more good idea to change colors

### DIFF
--- a/guake/boxes.py
+++ b/guake/boxes.py
@@ -122,7 +122,7 @@ class RootTerminalBox(Gtk.Overlay, TerminalHolder):
         self.search_frame.get_style_context().add_class("background")
         css_provider = Gtk.CssProvider()
         css_provider.load_from_data(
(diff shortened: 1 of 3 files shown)


Davidy22 (commented):
Just on a first impression from building and trying it out, the tabs are visibly compressed horizontally which I'm not quite sure I like, and the tab colors don't fill the full tab which makes it look like weird little flat squares in the middle of each tab. The flat color is a little jarring against the gradients that most themes apply to the tab bar, some transparency in the applied color box might be good. Clicking the cancel button breaks the dialog until the user opens the color picker dialog again.

selam:
you are right, coloring tabs looks very weird and i suspect themes couse of that, i can't figure out how to fix it,  i send this patch as a feature request and a example.

as a web developer i dont have enough experience and knowlage about gtk :(

PR #2242, opened 2024-04-10
Update prefs.py

novaTopFlex:

- rebase on latest HEAD:

  ```bash
  ```

- hack your change

  ```bash
  ```

  generation, using:

  ```bash
  ```

- create new commit message

  ```bash
  $ <hack the code>
  $ git commit --all
  ```

  line in your commit messages:

  ```
  # For a bug fix, uses:
  sem-ver: bugfix

  # For a new feature, uses:
  sem-ver: feature

  ```
This pull request is necessary to ensure that Python versions beginning with 3.10 are available as default shell options on the Guake terminal. Without this modification the newest Python version that would function on the Guake terminal is only Python 3.9, and Ubuntu 22.04 LTS (my host system) ships with Python 3.10 by default, causing the issue of Python 3.10 not appearing on my host system.

### DIFF
--- a/guake/prefs.py
+++ b/guake/prefs.py
@@ -64,7 +64,7 @@ log = logging.getLogger(__name__)
 
 # A regular expression to match possible python interpreters when
 # filling interpreters combo in preferences (including bpython and ipython)
-PYTHONS = re.compile(r"^[a-z]python$|^python\d\.\d$")
+PYTHONS = re.compile(r"^[a-z]python$|^python\d\.\d$|^python\d\.\d\d$")
 
 # Path to the shells file, it will be used to start to populate
 # interpreters combo, see the next variable, its important to fill the


novaTopFlex (commented):
By changing the regex for the Python interpreter listings, versions of the Python 3.x interpreter newer than late-2021 should be permitted as options for the Guake drop-down terminal. The command for "python3.10" as well as commands for newer versions of Python 3 violate the rules created by the old regex (from the first commit). This second commit should enable the "ipython3" and legacy "ipython2" commands on Linux distributions that differentiate between Python 2 and Python 3 versions of the IPython interpreter. Also, on various Ubuntu versions, the "ipython3" command is the IPython command for default Python 3 interpretation.

Davidy22 (commented):
Sorry about the late response, been busy as of late.

Few notes, you can use {x,y} notation for a range of repetitions instead of doing the two copied out repetitions of the last case, and it makes extending the regex a little more elegant too in case python ever gets into triple digit territory for the 3.x line. Also the ipython case seems to match a lot more letters than just i, is there a reason why we want to also match all these other cases like gpython and tpython?

SparkyBluefang:
This might be a simpler regex to evaluate
```python
PYTHONS = re.compile(r"^[a-z]?python(\d(\.\d+)?)?$")
```
or
```python
PYTHONS = re.compile(r"^[a-z]?python(?:\d(?:\.\d+)?)?$")
```
if you don't need to capture the version components.

PR #2257, opened 2024-09-08
Makefile: ensure that schema source files are in place before schema recompilation occurs

### LINKED ISSUE #2219: Makefile: race condition between install-schemas and compile-schemas
**Describe the bug**

The Makefile contains an [`install-system` target](https://github.com/Guake/guake/blob/47312dc2481b2fa617afe79cdcfddadf60e84516/Makefile#L71) that depends on both `install-schemas` and `compile-shemas` (sic).

Hoever: the `compile-shemas` target reads files from paths that are created by the `install-schemas` target, and they can both run in parallel when `make`'s parallelism is enabled (`-j`).

**Expected behavior**

The `.gschema.xml` files required for the `compile-shemas` target should always exist before the `compile-shemas` target is run.

**Actual behavior**

When `make` parallelism is enabled a `gschemas.compiled` file may or may not be created, depending on the runtime behaviour of the race condition.

**To Reproduce**

The sample below uses `make -j12 install-system` to approximate the behaviour:

```sh
$ PREFIX=/.../bugreport/ make -j12 install-system
install -dm755                                       "//.../bugreport//share/applications"
if [ 1 = 1 ]; then glib-compile-schemas //.../bugreport//share/glib-2.0/schemas; fi
for f in $(find po -iname "*.mo"); do \
        l="${f%%.*}"; \
        lb=$(basename $l); \
        install -Dm644 "$f" "//.../bugreport//share/locale/$lb/LC_MESSAGES/guake.mo"; \
done;
install -Dm644 "/.../guake/data/guake.desktop"       "//.../bugreport//share/applications/"
# you probably want to execute this target with sudo:
Error opening directory “//.../bugreport//share/glib-2.0/schemas”: No such file or directory
make: *** [Makefile:147: compile-shemas] Error 1
```

Note: sometimes the output differs, due to the race condition; for example if the directory has been created but does not yet contain `.gschema.xml` files then a message `No schema files found: doing nothing.` appears instead.

jayaddison:
When `make` is invoked in parallel mode (`-j` parameter), the dependencies of an individual makefile target may be resolved in unpredictable order.

That meant that sometimes, `compile-schemas` would incorrectly begin to run before all of the `gschema` XML files had been written to the schema destination directory.

This change relocates the `glib-compile-schemas` step to ensure that it occurs after the schema files have been written.

In addition, the existing `SCHEMA_DIR` variable is used instead of `gsettingsschemadir` to refer to the schema directory.

Resolves #2219

Edit: 20240914: redraft this description for brevity.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,7 +126,6 @@ uninstall-dev-locale:
 	@rm -rf guake/po
 
 install-schemas:
-	if [ $(COMPILE_SCHEMA) = 1 ]; then glib-compile-schemas $(DESTDIR)$(SCHEMA_DIR); fi
 	install -dm755                                       "$(DESTDIR)$(datadir)/applications"
 	install -Dm644 "$(DEV_DATA_DIR)/guake.desktop"       "$(DESTDIR)$(datadir)/applications/"
 	install -Dm644 "$(DEV_DATA_DIR)/guake-prefs.desktop" "$(DESTDIR)$(datadir)/applications/"
@@ -143,6 +142,7 @@ install-schemas:
 	install -Dm644 "$(DEV_DATA_DIR)"/*.glade "$(DESTDIR)$(GLADE_DIR)/"
 	install -dm755                                         "$(DESTDIR)$(SCHEMA_DIR)"
 	install -Dm644 "$(DEV_DATA_DIR)/org.guake.gschema.xml" "$(DESTDIR)$(SCHEMA_DIR)/"
+	if [ $(COMPILE_SCHEMA) = 1 ]; then glib-compile-schemas $(DESTDIR)$(SCHEMA_DIR); fi
 
 uninstall-system: uninstall-schemas uninstall-locale
 	$(SHELL) -c $(PYTHON_SITEDIRS_FOR_PREFIX) \
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,7 @@ ln-venv:
 clean-ln-venv:
 	@rm -f .venv
 
-install-system: install-schemas compile-shemas install-locale install-guake
+install-system: install-schemas install-locale install-guake
 
 install-guake:
 	# you probably want to execute this target with sudo:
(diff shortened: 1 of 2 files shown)


jayaddison:
Ok, my apologies: I definitely should have tested this before opening a pull request.

From attempting the build process locally: it seems that in fact the `install_schemas` step _is_ intended to occur before `compile_schemas` -- the latter requires that the files have already been written to the `$(SCHEMA_DIR)` directory.

I've moved this back into 'draft' pull request status while confirming more of the details.

jayaddison:
It probably is worth adding a `reno` entry for this after all.. this race condition can be a source of build nondeterminism, and that means that some downstream packagers might have added `-j1` or potentially even infrastructure-level workarounds to prevent `guake` building in parallel environments.  So an informational message in the changelog could help as a hint that it should be OK to remove those workarounds.

jayaddison:
cc @Davidy22 - I made a bit of a mess developing this PR - my apologies for not testing this until _after_ I opened it - but I believe that it is ready now and resolves a genuine build-time race condition for `guake`.

Davidy22:
Sorry about the delay in checking the build, and seems like github did some deprecating in the time between your testing and me checking this. I could do that seperately and you could pull, or you could bump the version for the `upload-artifact` action in your tree to v4

jayaddison:
Thanks for running the CI checks and taking a look!  I generally prefer to keep changes like those separate, so if that's OK with you, I'll wait until that is resolved and then will rebase/pull.

Davidy22:
Alright, there's still some things to resolve with CI but it should be in a state where you can update your branch and checks here should pass.

Davidy22 (approved):
Alright, I was fairly sure the visible change wasn't going to break anything, CI results agree, merging.

jayaddison:
Somewhat-delayed reply: thank you @Davidy22!

Issue #2019, opened 2022-01-22
Config option for window dimensions in pixels


cowboycodur:
### Feature 
Ability to toggle on and off specifying window height and width  in pixels instead of a percentage.

### Use Case
If you have monitor A at 5120x1440 and Monitor B at 1920x1080, when using monitor A full width is very unwieldy so if you set it to 30%, when switching to monitor B the width is too small for use.  If you could set it to always be 1920 it would work on both monitors. This would be very useful for laptops when docked vs undocked.

Dkacurik:
Hey, I'd like to work on this issue. Can I get started?

Davidy22:
Yeah sure, submit a PR for it when you've got it done

amontano495:
Hello I would like to work on this issue.

PR #2267, opened 2024-10-22
Terminal window configured by pixel size

adamm-xyz:
This change will allow the user to configure the size of the terminal window with pixel dimensions rather than solely using the sliding scales for height and width. The scale and spin box widgets are updated to reflect each other's change real time. A few new changes needed to be made to the settings schema in order for this to work. 

Also this should fix issue #2019

### DIFF
--- a/guake/prefs.py
+++ b/guake/prefs.py
@@ -1182,7 +1182,9 @@ class PrefsDialog(SimpleGladeApp):
         self.get_widget("window_width").set_value(value)
 
         value = self.settings.general.get_int("window-pixel-height")
+        max_value = self.settings.general.get_int("max-window-pixel-height")
         self.get_widget("window_vertical_dimension").set_value(value)
+        self.get_widget("window_vertical_dimension").set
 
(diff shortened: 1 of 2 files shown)


Davidy22 (commented):
Sorry it took me so long to get back to you on this, been a little busy lately.


Currently this PR seems to default to using 100x100 pixel dimensions, which is going to be jarring for the user. Probably would be better to set the pixel dimensions from the existing percentages and monitor dimensions for first run instead of setting default starting dimensions to 100x100. Since you have minimums set maybe can set the pixel width/height to 0/0 at start and if the pixel dimensions are out of range, which they will be if you set them to be 0/0 initially, then calculate pixel dimensions.

Alternately, we could just skip saving seperate values for pixel dimensions entirely and do the scale calculation every time we populate/use the spin box. That'd save on stuff like the redundant size setting that would make us have to do extra logic in `load_configs()` and `set_final_window_rect()` and we can cut some extra settings in gschema too.

Issue #2238, opened 2024-04-03
Allow specifying the number of equal parts to split the terminal tab into


givascu:
As an user, I would like to be able to split the terminal tab into more than two equal parts, e.g. split the terminal vertically into three parts with equal width. Currently Guake only allows splitting in half.

To do this, I propose adding an optional parameter to the `--split-vertical` and `--split-horizontal` command line options to specify the number `N` of equal parts to split into, e.g.

```
  --split-vertical   [N]   Split the selected tab vertically into N equal parts (default: 2).
  --split-horizontal [N]   Split the selected tab horizontally into N equal parts (default: 2).
```

Iam-samyog:
Hey! Can I work on this?

Manyy8:
Hi! I’d like to work on this issue. I can implement support for specifying the number of equal parts with --split-vertical N and --split-horizontal N, along with the necessary validation, tests, and documentation.

If this issue is still available, I’d be happy to take it up. Thanks!

PR #2317, opened 2026-02-17
Added functionality for multiple equal tabs.  Regarding Issue #2238

lucasperez-techdev:
#2238 
Please review my code changes for this issue. Added 2 new arguments, --split-vertical-tabs and --split-vertical-tabs. Both of these arguments take in a value from 2 and greater and splits the terminal selected into an N number of equally split terminals.


Davidy22:
The change seems good, but the name had me under the impression that this did something with the things that are already called tabs. Can the name not be panels or something? Also, a release note file should be made for this, do a `make reno slug` and fill the file in.

PR #2323, opened 2026-04-23
Fix window width on multi-monitor setups with different resolutions

nox-404:
## Summary

On multi-monitor setups with different widths, Guake would sometimes open or resize to the wrong width.

## Tab spawn issue

When creating a new Tab, there would be a high chance of the window width not matching the current monitor

## Hide/Show issue

When hiding and showing the window, it would sometimes shows with a width not matching the current monitor

## Test

I manually tested this issue on a multi-monitor setup (1920x1200 and 2560x1440).

  - WM/Compositor: **Sway** `1.11`
  - Session: **Wayland** `(GDK_BACKEND=wayland)`
  - GTK: `3.24.52`
  - VTE `0.84.0`

sem-ver: bugfix

### DIFF
--- a/guake/guake_app.py
+++ b/guake/guake_app.py
@@ -822,7 +822,7 @@ class Guake(SimpleGladeApp):
         self.settings.general.triggerOnChangedValue(
             self.settings.general, "window-ontop", user_data=user_data
         )
-        if not self.fullscreen_manager.is_fullscreen():
+        if not self.fullscreen_manager.is_fullscreen() and not terminal_uuid:
             self.settings.general.triggerOnChangedValue(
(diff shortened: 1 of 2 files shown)


nox-404:
I'm no python expert so feel free to tell me if something isn't right or if it might not work on other environment. I did get help from `Claude` to find and fix that issue :pray:

Davidy22:
`get_window()` was introduced a long time before `move_resize()`, which can lead to undefined errors for people on old systems. Just checking that `get_window()` returns is insufficient protection. If using move_resize() resolves the issue for you, we probably should still look into this though, probably should be a version check.

PR #2327, opened 2026-06-02
Add terminal pane reordering

peroxy:
Adds terminal pane reordering for split terminal panes through local hotkeys and context menu actions. The pane move commands reuse directional pane lookup and swap terminal boxes while preserving split positions.

Validation:
- `make reno-lint`
- `python3 -m pytest guake/tests/test_menus.py guake/tests/test_split_utils.py`


Davidy22:
Looks like black took issue with the new test cases, just run a quick pass of that, looks good otherwise.

PR #2329, opened 2026-06-20
Add tab activity highlighting with grace periods

zhum:
Highlight the title of a background tab (bold + orange) when its terminal produces output, clearing once the tab is selected. Gated behind a new "display-tab-activity" setting, off by default. To avoid lighting up every tab on session restore, activity is ignored for adjustable grace periods: a new-tab grace (default 3s) starting at tab creation, and a shorter focus-loss grace (default 1s) when a tab loses focus.


Davidy22:
Changes look fine, CI error is a stale thing on our end. Just need to do a little release note entry with a `make reno SLUG=[your text here]`

zhum:
Added release note.

PR #2331, opened 2026-07-06
Fix tab shortened path #2321

### LINKED ISSUE #2321: Shortened path in tab label works only for the first tab
As said in the title...

Image

```
$ guake --version
Guake Terminal: 3.11.dev0
VTE: 0.82.3
VTE runtime: 0.82.3
Gtk: 3.24.51
```

RarityFan437:
Fixes the layout issue where tab titles failed to update when multiple tabs or splits were active. Replaced direct terminal parent fetching with an iterative tree traversal to correctly find the notebook page index.

Closes #2321

--
RarityFan437

sem-ver: bugfix

### DIFF
--- a/guake/guake_app.py
+++ b/guake/guake_app.py
@@ -1125,9 +1125,13 @@ class Guake(SimpleGladeApp):
 
         # TODO NOTEBOOK this code only works if there is only one terminal in a
         # page, this need to be rewritten
-        for terminal in self.get_notebook().iter_terminals():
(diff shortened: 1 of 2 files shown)


Davidy22 (changes requested):
Not strictly a part of your changes, but the comment immediately above the changed block does quite seem like a comment that will become less accurate with this change and probably should be changed too.

RarityFan437:
@Davidy22 Done.

PR #2332, opened 2026-07-17
feat(prefs): add word characters preference

### LINKED ISSUE #2304: add editbox to tweak word-chars from preferences
Hi there,
I have no way of fiddling with the word-chars, as they do not appear in the "Preferences" windows.
It would also be nice to export/import any changes in word-chars when --save-preferences / --load-preferences is used.

in the meantime, please suggest an way to tweak the word-char sequence.

Thank you

royy92:
### What

Add a new Preferences field for configuring the `word-chars` setting.

### Why

Issue #2304 points out that users currently have no way to configure the `word-chars` value from the Preferences window.

This also makes the setting easier to manage when using `--save-preferences` and `--load-preferences`.

### How

- Add a new `word_chars_entry` field to `guake/data/prefs.glade`.
- Load, update, and reset the `word-chars` setting in `guake/prefs.py`.
- Add a release note in `releasenotes/notes/word-chars-preference-9fc892802ca01441.yaml`.

### Testing

- Ran `make style`.
- Ran `make check`.
- Ran `make test` (30 tests passed).
- Verified that changing the field updates the `word-chars` setting.
- Verified that resetting Compatibility preferences restores the default `word-chars` value.

### Screenshot

The new **Word characters** preference available in the **Compatibility** section.

Word characters preference

Closes #2304

### DIFF
--- a/guake/data/prefs.glade
+++ b/guake/data/prefs.glade
@@ -3597,10 +3597,29 @@ For example, use &lt;b&gt;subl %(file_path)s:%(line_number)s&lt;/b&gt; for Subli
                                   </packing>
                                 </child>
                                 <child>
-                                  <placeholder/>
+                                  <object class="GtkLabel" id="word_chars_label">
+                                    <property name="visible">True</property>
+                                    <property name="can-focus">False</property>
+                                    <property name="label" translatable="yes">Word _characters:</property>
+                                    <property name="use-underline">True</property>
+                                    <property name="xalign">0</property>
+                                    <property name="mnemonic-widget">word_chars_entry</property>
+                                  </object>
+                                  <packing>
+                                    <property name="left-attach">0</property>
+                                    <property name="top-attach">2</property>
+                                  </packing>
                                 </child>
                                 <child>
-                                  <placeholder/>
+                                  <object class="GtkEntry" id="word_chars_entry">
+                                    <property name="visible">True</property>
+                                    <property name="can-focus">True</property>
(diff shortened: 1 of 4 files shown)


guake/data/prefs.glade
```diff
@@ -3597,10 +3597,29 @@ For example, use &lt;b&gt;subl %(file_path)s:%(line_number)s&lt;/b&gt; for Subli
                                   </packing>
                                 </child>
                                 <child>
-                                  <placeholder/>
+                                  <object class="GtkLabel" id="word_chars_label">
+                                    <property name="visible">True</property>
+                                    <property name="can-focus">False</property>
+                                    <property name="label" translatable="yes">Word _characters:</property>
```
Davidy22: Why the underscore in the middle?

Davidy22 (changes requested):
Did a test to make sure it works, noticed that it doesn't immediately apply, only on next startup. May be unintuitive.

I was going to comment on the change to the comment as well but the shortened version grew on me on the second reading. Not sure why you did the label like that though, there are others with underscores but we have them conventionally at the front not middle.

royy92:
Thank you for catching that! I updated `word-chars` so changes are now applied immediately to existing terminals without requiring a restart. I also moved the mnemonic underscore to the beginning of the label to match the existing convention. I tested changing, clearing, and resetting the value without restarting Guake.

Davidy22 (approved):
Looks good, merging.

PR #2334, opened 2026-07-18
Add activity cooldown indication

zhum:
Added new tab indication: if there was activity, but then it stopped and no changes happened for 30 seconds (may be changed in settings), tab color is changed to blue.


guake/data/org.guake.gschema.xml
Davidy22: There is a mental/usability cost to every setting that we put in the preferences menu, these seem like settings that people aren't going to actually touch. Like I'd say the colors would be more likely to be touched than these times, and even then probably unlikely. Can just trim it down to the checkbox.

guake/data/prefs.glade
Davidy22: See previous note

guake/boxes.py
```diff
@@ -640,14 +640,17 @@ def remove_dead_child(self, child):
 # Foreground color applied to a background tab's title when it has unseen
 # activity. Chosen to stay readable on both light and dark tab bars.
 TAB_ACTIVITY_COLOR = "#E8A33D"
```
Davidy22: Usually constants like this should go at the top of the file, but also these are only getting used once so they may be better named inside `_render()` where they're used.

guake/boxes.py
```diff
@@ -676,22 +679,39 @@ def set_activity(self, active):
...
     def get_activity(self):
-        return self._activity
+        return self._activity_state is not None
+
+    def get_activity_state(self):
+        return self._activity_state
 
     def _render(self):
-        if self._activity:
+        if self._activity_state == "active":
```
Davidy22: Generally better to use enums for nonbinary states instead of strings

guake/notebook.py
```diff
@@ -450,9 +452,46 @@ def on_terminal_activity(self, terminal):
...
+        if hasattr(label, "set_activity_state"):
+            label.set_activity_state("active")
+        self._ensure_activity_timer()
+
+    def _ensure_activity_timer(self):
+        """Start the periodic tick that fades ACTIVE tabs to STALE once their
+        cooldown elapses. A single shared timer is used regardless of how
+        many tabs are currently highlighted."""
+        if self._activity_timer_id is None:
+            self._activity_timer_id = GLib.timeout_add(1000, self._tick_activity_states)
```
Davidy22: Why are we doing it this way? The timeout call lets you just set the time. Is it to reduce tick calls? I mean, with a 30 second cooldown you'd need to have stuff happening in more than 30 tabs for this to be savings.

releasenotes/notes/bugfix-503a9fee8f80e714.yaml
Davidy22, line 5: Don't touch unrelated reno files even if you're hardly changing anytthing

releasenotes/notes/tab_activity_cooldown-5cf02db0a9a035dd.yaml
```diff
@@ -0,0 +1,25 @@
+release_summary: >
```
Davidy22, line 1: Put everything you want to say in one reno file

scripts/bootstrap-dev-pip.sh
Davidy22, line 11: What on earth are you doing here?

Davidy22 (changes requested):
Sorry about the delay, been busy. There's been a fair few changes, you'll have to rebase on top of the current master. Also some strange meta changes that you shouldn't need to do here, and one baffling change. If you're involving AI, it would be much preferred if you could also review the output the AI is giving you

scripts/bootstrap-dev-pip.sh
zhum: Otherwise it doesn't work even in venv for my python 3.12.3...

releasenotes/notes/tab_activity_cooldown-5cf02db0a9a035dd.yaml
zhum: Done

guake/data/prefs.glade
zhum: Fixed

zhum (commented):
Updated PR according to the comments.

PR #2335, opened 2026-07-19
fix: disable XKB cooked accelerators to restore F12 global hotkey

Twilight0:
The keybinder library's XKB modifier computation (FinallyGetModifiersForKeycode) spuriously returns Mod1/Alt as a consumed modifier for F12 at level 0 on certain systems (e.g. Cinnamon/Muffin). This causes XGrabKey to attempt grabbing Alt+F12 instead of plain F12, which fails because the X server reserves Alt+Fn keys for VT switching.

Disabling cooked accelerators matches the approach used by Tilda terminal's tomboykeybinder.c, which does not use XKB at all for modifier computation. Users who need cooked accelerators can re-enable them via GSettings.

sem-ver: bugfix

### DIFF
--- a/guake/guake_app.py
+++ b/guake/guake_app.py
@@ -265,6 +265,12 @@ class Guake(SimpleGladeApp):
         # loading and setting up configuration stuff
         GSettingHandler(self)
         Keybinder.init()
+        # Disable XKB cooked accelerators to work around a keybinder
+        # library bug where FinallyGetModifiersForKeycode() returns
+        # spurious consumed modifiers for function keys (e.g. Alt for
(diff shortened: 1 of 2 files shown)


isabekov:
Thanks for the patch! It fixed the issue with unresponsive F12 shortcut on Arch Linux (kernel version 7.1.5)  running `guake v3.10.1`.

Twilight0:
Well, I suppose someone has to merge it already!

anhtumai (approved):

Davidy22:
Looks good to me, merging.

Issue #2337, opened 2026-07-26
Feature Request: Allow resizing the drop-down terminal height by dragging its bottom edge


anhtumai:
Context: Similar to how Windows Terminal allows it, I'd like to click and drag the bottom bar of the drop-down terminal up or down to adjust its height. This is especially useful when running commands with long output, as you can temporarily expand the terminal to see more without leaving the drop-down view.

Current behavior: The drop-down terminal has a fixed height.
Expected behavior: A draggable resize handle on the bottom edge.

NoumanParvez12:
@anhtumai  i would like to work on it

anhtumai:
@NoumanParvez12  I have created a PR already: https://github.com/Guake/guake/pull/2343 You can take a look and give feedback

NoumanParvez12:
@anhtumai oh sorry, I am new to this and didn't know that

PR #2338, opened 2026-08-02
2337 allow resize the drop down terminal

anhtumai:

- rebase on latest HEAD:

  ```bash
  ```

- hack your change

  ```bash
  ```

  generation, using:

  ```bash
  ```

- create new commit message

  ```bash
  $ <hack the code>
  $ git commit --all
  ```

  line in your commit messages:

  ```
  # For a bug fix, uses:
  sem-ver: bugfix

  # For a new feature, uses:
  sem-ver: feature

  ```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,15 +8,15 @@ repos:
           - "--markdown-linebreak-ext=md"
         exclude: COPYING
   - repo: https://github.com/PyCQA/flake8.git
-    rev: 3.9.2
+    rev: 7.3.0
     hooks:
       - id: flake8
   - repo: https://github.com/PyCQA/pylint.git
-    rev: v2.13.6
+    rev: v3.3.9
     hooks:
       - id: pylint
   - repo: https://github.com/psf/black.git
-    rev: 22.10.0
+    rev: 24.10.0
     hooks:
       - id: black
   - repo: https://github.com/hackedd/fiximports.git
--- a/guake/data/guake.glade
+++ b/guake/data/guake.glade
@@ -48,6 +48,7 @@
     <property name="skip-taskbar-hint">True</property>
     <property name="skip-pager-hint">True</property>
     <property name="urgency-hint">True</property>
+    <property name="decorated">False</property>
     <child>
       <object class="GtkBox" id="mainframe">
         <property name="visible">True</property>
--- a/guake/data/guake.glade
+++ b/guake/data/guake.glade
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.18.3 -->
+<!-- Generated with glade 3.40.0 -->
 <interface domain="guake">
   <requires lib="gtk+" version="3.10"/>
   <object class="GtkMenu" id="tray-menu">
     <property name="visible">True</property>
-    <property name="can_focus">False</property>
+    <property name="can-focus">False</property>
     <child>
       <object class="GtkImageMenuItem" id="prefs-menuitem">
         <property name="label">gtk-preferences</property>
         <property name="visible">True</property>
-        <property name="can_focus">False</property>
-        <property name="use_underline">True</property>
-        <property name="use_stock">True</property>
+        <property name="can-focus">False</property>
+        <property name="use-underline">True</property>
+        <property name="use-stock">True</property>
         <signal name="activate" handler="show_prefs" swapped="no"/>
       </object>
     </child>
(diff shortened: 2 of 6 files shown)


anhtumai:
# Allow resizing the drop-down terminal by dragging its bottom edge
closes #2337

## Summary

Similar to Windows Terminal, this lets you resize Guake's drop-down window by dragging its bottom edge — handy for temporarily expanding the terminal to read long command output without leaving the drop-down view.

- Enable `decorated` on the main window (`guake.glade`) so the window manager provides a native border, which makes it grabbable/draggable on the bottom edge.

- Debounce `configure-event` on the main window (300ms) to detect when a drag-resize has settled, and record the resulting height/width as percentages of the monitor workarea (`Guake.on_window_configure` / `on_resize_debounced` in `guake_app.py`).

- Store that as an in-memory `SettingsOverride` (`guake/utils.py`) that `RectCalculator.set_final_window_rect` prefers over the persisted `window-height`/`window-width` gsettings values, without writing them back to gsettings — so the size holds for the rest of the session (surviving hide/show via F12) without permanently changing your configured default.

- If you explicitly change height/width/alignment via Preferences, `accel_increase_height`/`decrease_height`, or `dconf`, the override is dropped (`GSettingHandler.size_changed`/`alignment_changed` in `gsettings.py`) so the persisted setting takes priority again, as expected.

## Notes

- The Fluxbox `show()` workaround previously re-triggered the `window-height` changed signal to force a re-apply; that would now also clear the override, so it was changed to call `RectCalculator.set_final_window_rect` directly instead, preserving the original Fluxbox fix without disturbing the in-session override.

- Also bumped the `black`/`flake8`/`pylint` pre-commit hook revisions, which were badly out of date and crashing under modern Python (old `black` referenced ast.Str, removed in Python 3.12+).

## Testing

- `make style check test`
- Manually: drag the bottom edge to resize, hide/show (F12) and confirm the size sticks; then change the height via Preferences and confirm that takes over cleanly.

anhtumai:
Demo of my PR:

https://github.com/user-attachments/assets/347621bf-25fa-4046-ab74-bc892a78e8c2

Davidy22:
I notice from both my testing and your video that this change introduces an unnecessary titlebar. That's going to have to get cleaned up before I merge this.

anhtumai:
Thank you for your respone. I will look at it tonight or this weekend

anhtumai:
@Davidy22  I have fixed the code.
Here is the new demo , when I have removed the title bar. You can check in `guake.glade` that the code is same as before, except for glade version changed

https://github.com/user-attachments/assets/501e5e08-4e98-4253-b4af-38f78edae811

anhtumai:
About the title bar, actually I prefer to keep it, because it has options to: Move to monitor (Up/Down/Left/Right). So we can move the Guake terminal to other places, not stuck in one place. What do you think about this @Davidy22 ?

image

If you agree, I will keep the titlebar, since it improves user experience a lot

Another solution is to make another issue to fix: Move Guake terminal to wherever cursor last clicks, but it will be much harder to do than just enabling the titlebar by setting: decorated=True

guake/guake_app.py
```diff
@@ -190,6 +191,21 @@ def load_schema():
         self.mainframe = self.get_widget("mainframe")
         self.mainframe.remove(self.get_widget("notebook-teminals"))
 
+        # Window is undecorated (no WM title bar/borders), so provide a thin
+        # drag handle at the bottom edge to let users resize it manually.
+        # This is for issue: https://github.com/Guake/guake/issues/2337
```
Davidy22, line 196: Don't put issue numbers in comments, just describe what the code does.

guake/guake_app.py
```diff
@@ -223,6 +239,11 @@ def load_schema():
         # Debounce accel_search_terminal
         self.prev_accel_search_terminal_time = 0.0
 
+        # holds the GLib source id for the pending resize-debounce timeout
+        # used to capture resizing activities for https://github.com/Guake/guake/issues/2337
```
Davidy22, line 243: Similar deal

guake/data/guake.glade
```diff
@@ -1,70 +1,70 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.18.3 -->
+<!-- Generated with glade 3.40.0 -->
 <interface domain="guake">
   <requires lib="gtk+" version="3.10"/>
   <object class="GtkMenu" id="tray-menu">
     <property name="visible">True</property>
-    <property name="can_focus">False</property>
+    <property name="can-focus">False</property>
```
Davidy22, line 7: This is a convention change to go with the glade version bump I presume?

.pre-commit-config.yaml
```diff
@@ -8,15 +8,15 @@ repos:
           - "--markdown-linebreak-ext=md"
         exclude: COPYING
   - repo: https://github.com/PyCQA/flake8.git
-    rev: 3.9.2
+    rev: 7.3.0
```
Davidy22, line 11: Don't change our linter versions. Out of scope, do this in another PR.

anhtumai:
> About the title bar, actually I prefer to keep it, because it has options to: Move to monitor (Up/Down/Left/Right). So we can move the Guake terminal to other places, not stuck in one place. What do you think about this @Davidy22 ?
> 
> image
> 
> If you agree, I will keep the titlebar, since it improves user experience a lot
> 
> Another solution is to make another issue to fix: Move Guake terminal to wherever cursor last clicks, but it will be much harder to do than just enabling the titlebar by setting: decorated=True

@Davidy22 sorry for direct ping. But can you give your feedback here?

Davidy22 (changes requested):
Adding a title bar changes the program quite substantially, is out of scope for this PR, and jars with how it's supposed to be a panel that pulls down from the top.

Flagged things I noticed, also the temporary override mechanism's a point of discussion. Size changes not being reflected in the preferences menu is going to be a bit unintuitive. Also need to consider what merits usage of the temporary settings, why would dragging the window edge count for a temporary change, but CTRL+up/down doesn't?

I would lean heavy towards cutting the temporary settings mechanism and just having the window edge dragging alter the general window size setting. If you want to make a case for doing it this way, at the very least the window size keyboard shortcuts fulfil a very similar role to this and should also use the same mechanism.

anhtumai:
> Adding a title bar changes the program quite substantially, is out of scope for this PR, and jars with how it's supposed to be a panel that pulls down from the top.
> 
> Flagged things I noticed, also the temporary override mechanism's a point of discussion. Size changes not being reflected in the preferences menu is going to be a bit unintuitive. Also need to consider what merits usage of the temporary settings, why would dragging the window edge count for a temporary change, but CTRL+up/down doesn't?
> 
> I would lean heavy towards cutting the temporary settings mechanism and just having the window edge dragging alter the general window size setting. If you want to make a case for doing it this way, at the very least the window size keyboard shortcuts fulfil a very similar role to this and should also use the same mechanism.

Thank you . I will modify the logic to align with existing implementation

PR #2339, opened 2026-08-03
Consume Ctrl+click once Guake has handled it

Suor:
### The bug

`GuakeTerminal.button_press()` handles Ctrl+click (hyperlink, quick-open on selection, quick-open matcher) but returns `None`, so the event keeps propagating and VTE's default handler also forwards the click to the child process as a mouse report. Any mouse-aware application running in the terminal then acts on the very same click.

Where I hit it: I run [herdr](https://herdr.dev) So Ctrl+clicking a URL opens **two identical browser tabs**.

### The fix

Return `True` from the Ctrl+click branch, but only when one of its branches actually handled the event:

```python
        if event.button == 1 and (event.get_state() & Gdk.ModifierType.CONTROL_MASK):
            ...
            elif matched_string is not None and matched_string[0]:
                self._on_ctrl_click_matcher(matched_string)
            else:
                return False
            return True
```


Davidy22:
Looks good to me, merging. Sorry about the delay

PR #2340, opened 2026-08-04
Run the quick-open command in the active tab dir

Suor:
Allows command to work for relative paths, including partial ones.

`sem-ver: bugfix`


Davidy22:
The subprocess call change seems reasonable, but what's going on with the whole extra file?

Suor:
Gobbled up it somehow, removed it

Davidy22:
Alright seems good, merging.

